PR #28, opened 2021-01-17
London Class 7 - Kara Howard - JavaScript Core 1 - Week 1

kmh256:
# Your Details

- Your Name: Kara Howard
- Your City: London
- Your Slack Name: Kara Howard

# Homework Details

- Module: JavaScript Core 1
- Week: 1

### DIFF
--- a/extra/2-piping.js
+++ b/extra/2-piping.js
@@ -16,26 +16,31 @@
   the final result to the variable goodCode
 */
 
-function add() {
-
+function add(num1, num2) {
+    return num1 + num2;
 }
+console.log(add(1, 3));
 
-function multiply() {
-
+function multiply(num1, num2) {
+    return num1 * num2;
 }
+console.log(multiply(2, 3));
 
-function format() {
-
+function format(number) {
+    return `£${number}`;
 }
+console.log(format(16));
 
-const startingValue = 2
+const startingValue = 2;
 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode = 
+let badCode = add(startingValue, 10); badCode = multiply(badCode, 2); console.log(badCode = format(badCode)); // code all on one line is difficult to read
 
 /* BETTER PRACTICE */
-
-let goodCode = 
+// Comments can be added to explain each line and it is easier to read
+let goodCode = add(startingValue, 10); // function call to add 10 to startingValue and store result in goodCode
+goodCode = multiply(goodCode, 2); // function call to multiply value of goodCode by 2
+console.log(goodCode = format(goodCode)); // function call to format value of goodCode
 
 /* ======= TESTS - DO NOT MODIFY ===== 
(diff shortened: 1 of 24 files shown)


askinekincihf:
Well done Kara, you have done a great work! Your codes are readable and comments are also helpful to understand the way how you handle the exercises. I also like the way that you committed after each task which makes your commit history very professional.  In addition to that, It is the fact that you know how to use the back tick as the delimiter which is introduced by ES6.  I would only advise that you can modify Magic 8 Ball project with adding random string selection options from the array/arrays or with any other alternatives instead of  writing hard coding. 

Great job overall, :clap::clap::clap:.

kmh256:
Thank you Askin for the helpful feedback and advice, it is much appreciated! I will continue to make improvements on my Magic 8 Ball solution :)

extra/3-magic-8-ball.js
```diff
@@ -103,22 +104,21 @@ console.log(shakeBall());
 
 function checkAnswer(answer, answerSelection) {
   //Write your code in here
-  let rating = "";
-   // if the "veryPositive" key values in the possibleAnswers object include the string assigned to "answer" at the index specified by "answerSelection", then return "rating" value very positive
+   // if the "veryPositive" values in the possibleAnswers object include the string assigned to "answer" at the index specified by "answerSelection", then return "very positive"
   if (possibleAnswers.veryPositive.includes(answer, [answerSelection])) {
```
askinekincihf, line 108: well done Kara. you used includes method as exactly as described in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes.  Your codes are brilliant!
:clap::clap:

kmh256: Thank you for all your helpful suggestions :)

extra/2-piping.js
```diff
@@ -16,26 +16,31 @@
...
-let badCode = 
+let badCode = add(startingValue, 10); badCode = multiply(badCode, 2); console.log(badCode = format(badCode)); // code all on one line is difficult to read
 
 /* BETTER PRACTICE */
-
-let goodCode = 
+// Comments can be added to explain each line and it is easier to read
+let goodCode = add(startingValue, 10); // function call to add 10 to startingValue and store result in goodCode
+goodCode = multiply(goodCode, 2); // function call to multiply value of goodCode by 2
+console.log(goodCode = format(goodCode)); // function call to format value of goodCode
```
builtbybea, line 43: Great work Kara ✨ One suggestion rather than reassigning `goodCode` when you want to call another function, you can  nest function calls for example:
`let goodCode = format(multiply(add(startingValue, 10), 2));`
This still gives the same result but this is an alternative solution and is all done to preference. 

Keep up the good work

builtbybea (commented):
Excellent work on this coursework! Keep up the good work Kara ✨ ✨

PR #30, opened 2021-01-18
London Class 7 - Yunus FIRAT - JavaScript Core 1 - Week 1

yunusfirat:
# Your Details

- Your Name:Yunus FIRAT
- Your City:London
- Your Slack Name:yunusfirat44

# Homework Details

- Module:javaScript core
- Week:1

### DIFF
--- a/mandatory/4-tax.js
+++ b/mandatory/4-tax.js
@@ -5,7 +5,12 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  let totalPriceWithTax = price + price / 5;
+  
+  return totalPriceWithTax;
+
+}
 
 /*
   CURRENCY FORMATTING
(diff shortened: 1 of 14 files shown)


mandatory/4-tax.js
```diff
@@ -24,10 +22,8 @@ function calculateSalesTax(price) {
 
 function addTaxAndFormatCurrency(price) {
   let totalPrice = calculateSalesTax(price);
-  let decimalNumberOfAbove = Number(totalPrice).toFixed(2);
```
ademalagoz: Hi Yunus, great work, well done.
As a small suggestion: would be to use lowercase words for starting to write new variables.(eg:Number)

Jpsi:
Great submission, Yunus! 👍 I like, in particular, the names that you've given variables throughout your code - they make it easy to read.

mandatory/4-tax.js
Jpsi: > As a small suggestion: would be to use lowercase words for starting to write new variables.(eg:Number)

Here, `Number` is not a variable; rather, it's the [Number function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#description) that takes a non-number parameter and attempts to convert it to an object of type Number.

Note that using the Number function is not strictly needed here, as `totalPrice` is already a number.

PR #32, opened 2021-01-18
London Class 7 - Ionut Tripa - Javascript Core 1 - Week 1

iTrip23:
# Your Details
Ionut Tripa
London
Johnny

# Homework Details

Javascript Core 1
Week 1

### DIFF
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,3 +1,2 @@
-// Start by creating a variable `greeting`
-var greeting = 'Hello sunshine'
+var greeting = 'Hello sunshine';
 console.log(greeting);
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,3 +1,3 @@
 // Start by creating a variable `greeting`
-
+var greeting = 'Hello sunshine'
 console.log(greeting);
--- a/exercises/I-floats/exercise.js
+++ b/exercises/I-floats/exercise.js
@@ -1,8 +1,9 @@
-var numberOfStudents = 15;
-var numberOfMentors = 8;
+let numberOfStudents = 15;
+let numberOfMentors = 8;
+let total = numberOfMentors + numberOfStudents;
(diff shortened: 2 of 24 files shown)


exercises/C-variables/exercise.js
```diff
@@ -1,3 +1,3 @@
 // Start by creating a variable `greeting`
-
+var greeting = 'Hello sunshine'
```
Laura-TM, line 2: Perhaps you could add a (;) at the end of the variable

exercises/K-functions-parameters/exercise3.js
```diff
@@ -1,4 +1,5 @@
 // Write your function here
+const createGreeting = firstName => 'Hi ' + firstName + ', How are you?';
```
Laura-TM, line 2: Nice use of syntax and labels for variables

exercises/I-floats/exercise.js
```diff
@@ -1,2 +1,8 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+
+var percentageOfStudents = Math.round(15*100/23) + '%';
+var percentageOfMentors = Math.round(8*100/23) + '%';
+
+console.log(percentageOfStudents);
+console.log(percentageOfMentors);
```
oksygenn, line 8: Hi Johnny,
I have a little suggestion :)
You could add variables `numberOfStudents` and `numberOfMentors` and save the result to a new variable like `total` and then use all these variables in `Math.round()` instead of just using numbers(15, 8, 23).  Something like this: `Math.round(numberOfStudents*100/total)`.  
This will make your code less hard-coded, so you can change numberOfStudents or numberOfMentors to any other numbers - percentage would still be calculated correctly 😉

Also, I believe the expected result was "Percentage students: 65%
Percentage mentors: 35%".

oksygenn (commented):
Great job with the homework Johnny!
Loved the short functions :)
Only had one suggestion about hard-coded values.

PR #41, opened 2021-01-20
South Africa Class_1 - Jonathan Heugh - JavaScript-Core-1 - Week 1

Jontas77:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:


exercises/L-functions-nested/exercise2.js
```diff
@@ -0,0 +1,20 @@
...
+var mentor2 = "Irina";
+var mentor3 = "Mimi";
+var mentor4 = "Rob";
+var mentor5 = "Yohannes";
+
+function myGreeting (greet) {
+    return greet;
+}
+
+var result = myGreeting ('HELLO') + ' ' + mentor1.toUpperCase();
```
sirpfaira: You should have put these repeated statements in a method, because the method actually does nothing

ShawenHarker (commented):
Hi Jontas, 
Firstly I would like to say your code is super clean bud, well done on that man but on a more serious note, I would like to point your parameters for your functions, can you give it more descriptive names please, as "s" and "m" makes the code a bit harder to read, especially if you doing a quick scan through your code, other than that, well done man. 

Keep up the good work!!!

mandatory/3-function-output.js
```diff
@@ -1,55 +1,58 @@
...
-);
-test(
-  "concatenate function - case 3 works",
-  concatenate("I", "am", 13),
-  "I am 13"
-);
+// Add comments to explain what this function does. You're meant to use Google!
+// This function get a random number between 0 and 1 inclusive 0 but not 1 multiply by 10
+function getNumber() {
+  return Math.random() * 10;
```
Ntshembo-Hlongwane1, line 4: Hey there great stuff on the previous work **however**....

- So when you use ```Math.random() * 10``` this won't really guarantee giving you **whole** number 1-10 (Note the question was **ambiguous** as to are floating number required but if for instance **only whole numbers were needed**) this would not generate one **from what I know I stand to be corrected**

- To generate random whole number from 0-10  code would look like this below:
```js
Math.floor(Math.random() * 10)
```

Jontas77:
> Hi Jontas,
> Firstly I would like to say your code is super clean bud, well done on that man but on a more serious note, I would like to point your parameters for your functions, can you give it more descriptive names please, as "s" and "m" makes the code a bit harder to read, especially if you doing a quick scan through your code, other than that, well done man.
> 
> Keep up the good work!!!

Hi Shawen thank you for your review I will go back and make the necessary changes.

mandatory/3-function-output.js
Jontas77: Hi Ntshembo thank you for your review I will go back and make the necessary changes.

exercises/L-functions-nested/exercise2.js
Jontas77: Hi thank you for your review I will go back and make the necessary changes

ezekiel444 (commented):
Nice one. But everything seems so unreal with the consistent deletion of questions, instructions and even test cases then pasting of answers.

extra/2-piping.js
```diff
@@ -1,64 +1,67 @@
...
+
+function multiply(numA, numB) {
+    return numA * numB;
+}
+
+function format(numC) {
+    return '£' + numC.toString();
+}
+
+const startingValue = 2 + 10; let result = startingValue * 2; let formatResult = '£' + result;
```
ezekiel444, line 31: it is bad practice to define everything on a single line

ezekiel444 (commented):
extra-2-piping: "const startingValue = 2 + 10; let result = startingValue * 2; let formatResult = '£' + result"; bad practice to define everything in same line. Nice work.

PR #49, opened 2021-01-21
London Class 7 - Rafael Flores - JavaScript Core 1 - Week 1

rfloresram:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/exercises/G-numbers/exercise.js
+++ b/exercises/G-numbers/exercise.js
@@ -1,1 +1,11 @@
 // Start by creating a variables `numberOfStudents` and `numberOfMentors`
+var numberOfStudents = 15
+var numberOfMentors = 8
+var totalOf = 15 + 8;
+console.log("Number of students: " + numberOfStudents )
+console.log("Number of mentors: " + numberOfMentors)
+console.log("Total number of students and mentors: " + totalOf)
+
+
+
+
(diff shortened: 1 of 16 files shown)


exercises/I-floats/exercise.js
```diff
@@ -1,2 +1,4 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+let totalOf = numberOfStudents + numberOfMentors
+console.log(`Percentage students: ${Math.round((numberOfStudents / totalOf) * 100)}%\nPercentage mentors: ${Math.round((numberOfMentors / totalOf) * 100)}%`);
```
DanGhanbari, line 4: Good work Rafael, I suggest breaking your codes down into smaller pieces for better reading to prevent confusion. So it would be better if you declare more variables and the console.log it

exercises/G-numbers/exercise.js
```diff
@@ -1 +1,11 @@
 // Start by creating a variables `numberOfStudents` and `numberOfMentors`
+var numberOfStudents = 15
+var numberOfMentors = 8
```
bisioolajide, line 3: Great work Rafael, I would only make a  small comment around your use of semi-colons that are missing on a few lines.

exercises/G-numbers/exercise.js
```diff
@@ -1 +1,11 @@
...
+var totalOf = 15 + 8;
```
lpalma, line 4: Hey Rafael :)
Here we are repeating ourselves by using the same numbers we used in the variables `numberOfStudents` and `numberOfMentors`.
Is there a way we can avoid using these numbers again? (Hint: you solved it in the exercise **I-floats**)

PR #52, opened 2021-01-22
London Class 7 - Hozan Ali - JavaScript-Core-1-Coursework - Week 1

Hozan94:
Solving all question in the 'exercise' and 'mandatory' folders.

# Your Details

- Your Name: Hozan Ali
- Your City: London Class 7
- Your Slack Name: Hozan Ali

# Homework Details

- Module: javaScript-core-1
- Week: 1

### DIFF
--- a/exercises/F-strings-methods/exercise.js
+++ b/exercises/F-strings-methods/exercise.js
@@ -1,3 +1,6 @@
 // Start by creating a variable `message`
+var myName = "Hozan";
+var nameLength = myName.length;
+var message = "My name is " + myName + " and my name is " + nameLength + " characters long";
 
 console.log(message);
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,4 +1,4 @@
 const name = " Daniel  ";
-var message = "My name is " + name.trim() + " and my name is " + name.length + " characters long";
+var message = "My name is " + name.trim() + " and my name is " + name.trim().length + " characters long";
 
 console.log(message);
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,3 +1,4 @@
 const name = " Daniel  ";
+var message = "My name is " + name.trim() + " and my name is " + name.length + " characters long";
 
 console.log(message);
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -2,13 +2,13 @@
 function getNumber() {
   return Math.random() * 10;      
 }
-// The function above will return a random integer from 0 and 9, The random() method returns a random number from 0 (inclusive) up to but not including 1 (exclusive)
+// The function above will return a random floating point between 0 and 10 (not included), The random() method returns a random floating number from 0 (inclusive) up to but not including 1 (exclusive)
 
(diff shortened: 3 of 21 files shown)


exercises/F-strings-methods/exercise.js
```diff
@@ -1,3 +1,6 @@
 // Start by creating a variable `message`
+var myName = "Hozan";
+var nameLength = myName.length;
+var message = "My name is " + myName + " and my name is " + nameLength + " characters long";
```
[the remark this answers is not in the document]
Hozan94: Thanks for the feedback Omar, the documents you provided looks great, I already went through them and found some useful info, I want to also note that in some exercises I did apply template literals, while in others I left them to have string literlas just so I practice how to use them both. Many Thanks

exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,4 @@
 const name = " Daniel  ";
+var message = "My name is " + name.trim() + " and my name is " + name.length + " characters long";
```
Domenico-cyf, line 2: Very well done on your work Hozan, impressive!
only one very personal suggestion, you could have used a variable to hold the trimmed value and then use it. That would had simplify your code. Now is a bit compacted, which again, is a personal choice. I read it better on multiple lines.

mandatory/3-function-output.js
```diff
@@ -1,16 +1,19 @@
 // Add comments to explain what this function does. You're meant to use Google!
 function getNumber() {
-  return Math.random() * 10;
+  return Math.random() * 10;      
 }
+// The function above will return a random integer from 0 and 9, The random() method returns a random number from 0 (inclusive) up to but not including 1 (exclusive)
```
louisechow, line 6: Hi Hozan, when if you run `Math.random` what output do you get? If you multiple the result of `Math.random` with 10 does it result in an integer?

mandatory/3-function-output.js
```diff
@@ -1,16 +1,19 @@
...
 
 // Add comments to explain what this function does. You're meant to use Google!
 function s(w1, w2) {
   return w1.concat(w2);
 }
+// The concat() method above is used to join two or more strings, so the function will return a new string of w1 + w2.
```
louisechow, line 11: The function `s` only has 2 arguments, so you can't join more than 2 strings.

louisechow:
Great work Hozan! 👏

exercises/F-strings-methods/exercise2.js
Hozan94: Many thanks @Domenico-cyf for your feedback, I will consider such thing next time, as I was only trying to minimize my code.

mandatory/3-function-output.js
Hozan94: Hi @louisechow , many thanks for the feedback.
You are right, I should have worded my answer properly.
The `Math.random` method will return a floating point number between 0 and 1 (not included), and when it is multiplied by 10 we will still get a floating point number between 0 and 10 (not included).

mandatory/3-function-output.js
Hozan94: @louisechow you are right, I should be careful with how I write my answer, in that particual example it will only be used to concatenate the two parameters w1 and w2.

PR #59, opened 2021-01-22
London Class 7 - Negin Dn - Javascript - core1- coursework - Week1

NeginDn:
# Your Details

- Your Name: Negin Dn
- Your City: London
- Your Slack Name: Negin Dn

# Homework Details

- Module: JS-1
- Week: 1


Hozan94 (commented):
Hi @NeginDn , great work overall as you managed to successfully complete the exercises
 . I added some comments to see which area you can improve on.

exercises/L-functions-nested/exercise.js
```diff
@@ -3,3 +3,18 @@ var mentor2 = "Irina";
 var mentor3 = "Mimi";
 var mentor4 = "Rob";
 var mentor5 = "Yohannes";
+
+function uppercaseName(mentorName) {
+  return mentorName.toUpperCase();
+}
+
+function greetingMaker(mentorName) {
+  let greeting = "HELLO " + uppercaseName(mentorName);
```
Hozan94, line 12: I think, your function 'greetingMaker' can have two parameters, and the reason for that is what if we wanted to change the word 'HELLO' to something else later in future. And to do that you can write it like this :

```
function greetingMaker(capitalGreeting, name){
    let greeting = capitalGreeting.toUpperCase() + " " + name
    return greeting;
};
```

exercises/C-variables/exercise.js
```diff
@@ -1,3 +1,3 @@
-// Start by creating a variable `greeting`
-
-console.log(greeting);
+//* Print your `greeting` to the console 3 times
+let greeting = "Welcome to coding world";
+for (let i = 0; i < 3; i++) console.log(greeting);
```
Hozan94, line 3: At this stage there is no need to have a 'for' loop, it also won't answer the question as the code is not adding anything. To have three 'console.log()' , you just need write them like this:
console.log(greeting);
console.log(greeting);
console.log(greeting);

exercises/E-strings-concatenation/exercise.js
```diff
@@ -1,3 +1,6 @@
 // Start by creating a variable `message`
+let myName = "Negin";
+let greeting = "Hello, My name is ";
+let message = greeting + " " + myName;
```
Hozan94, line 4: In here, there is no need to have the ` "  " ` to add extra space as you have already added it after the word `is  "`

Hozan94 (commented):
Hi @NeginDn , great work overall as you managed to successfully complete the exercises
 . I added some comments to see which area you can improve on.

exercises/L-functions-nested/exercise.js
NeginDn: You're absolutely right. I will change it. 
Thank you for the feedback

exercises/F-strings-methods/exercise.js
```diff
@@ -1,3 +1,5 @@
 // Start by creating a variable `message`
```
DibaKarimi, line 1: Hi Negin, Well done!
Your code is correct, also there is  another way you can try, it might be easier when using more than one variables.
let message= `Hello my name is ${myName} and my name is ${lengthName} character long`;

louisechow: Somayeh's suggestion is a good one to make your code more simple. The method is called string interpolation. https://www.freecodecamp.org/news/javascript-string-format-how-to-use-string-interpolation-in-js/

Be careful, when you use string interpolation in Javascript you have to wrap the string in backticks(`),  not single quotes (').

exercises/K-functions-parameters/exercise5.js
```diff
@@ -1,5 +1,14 @@
...
+  return (
+    "Hello, my name is " + personName + " and I'm " + personAge + " years old"
+  );
+}
 
-const greeting = createLongGreeting("Daniel", 30);
+let greeting = createLongGreeting("Daniel", 30);
+console.log(greeting);
 
+ greeting = createLongGreeting("Negin", 32);
```
louisechow, line 10: Small thing.. the indenting on line 10 and 13 is off.

extra/1-currency-conversion.js
```diff
@@ -15,7 +17,10 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(poundPrice) {
+let realPrice = poundPrice * 5.7 * 0.99;
```
louisechow, line 21: Could you indent the code in this function please?

louisechow:
Excellent job Negin! 🌟

PR #65, opened 2021-01-22
London 7/ahmad janah

AhmadJanah:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,3 +1,9 @@
 const name = " Daniel  ";
 
-console.log(message);
+let firstName = "Ahmad";
+
+let nameLength = firstName.length;
+
+let message = "My name is " + firstName + " and my name is " + nameLength + " characters long";
+
+console.log(message.trim());
--- a/exercises/I-floats/exercise.js
+++ b/exercises/I-floats/exercise.js
@@ -1,2 +1,8 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+
+let studentPercent = (numberOfStudents * 100)/(numberOfStudents + numberOfMentors);
+let mentorsPercent = (numberOfMentors * 100)/(numberOfStudents + numberOfMentors);
+
+console.log("Percentage students:", Math.round(studentPercent));
+console.log("Percentage mentors:", Math.round(mentorsPercent));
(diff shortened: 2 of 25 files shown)


exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,9 @@
 const name = " Daniel  ";
 
-console.log(message);
+let firstName = "Ahmad";
```
NeginDn, line 3: Your name variable should have some extra spaces here, so when you use trim, It has something to remove.

exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,9 @@
...
+
+let nameLength = firstName.length;
+
+let message = "My name is " + firstName + " and my name is " + nameLength + " characters long";
+
+console.log(message.trim());
```
NeginDn, line 9: After you added some extra spaces to your name, you can use the trim() on the variable itself, this way it will remove the extra spaces where it should

exercises/I-floats/exercise.js
```diff
@@ -1,2 +1,8 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+
+let studentPercent = (numberOfStudents * 100)/(numberOfStudents + numberOfMentors);
+let mentorsPercent = (numberOfMentors * 100)/(numberOfStudents + numberOfMentors);
+
+console.log("Percentage students:", Math.round(studentPercent));
```
NeginDn, line 7: Very well and readable, the only comment I can give is if you ad the % sign to the final result, it will be more like expected result

NeginDn (commented):
Hi @AhmadJanah  great work, you managed to successfully complete the exercises with very tidy and readable codes. I added some suggestion, have a look to see if they can help you

AhmadJanah:
Thank you @NeginDn for these very useful comments.
It is really appreciated.

mandatory/2-logic-error.js
```diff
@@ -1,18 +1,18 @@
...
+  return word.length;
 }
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  return a * b * c;
 }
 
+const { strict } = require("assert");
```
Jpsi, line 15: This line loads the module `assert` into the variable `strict`, but it looks like the latter is not used in the code. I would remove the line to make the code cleaner.

Jpsi (commented):
Great work - well done for completing so many exercises! 👍

I would provide more descriptive names for the commits (e.g.: "Complete mandatory exercises", or "Complete extra exercises"), as this gives anyone who is looking at the code a better idea of what the purpose of a commit is.

PR #74, opened 2021-01-23
London Class 7 - Joy Omodiale - JavaScript - Week 1

jmodiale:
# Your Details

- Your Name: Joy Omodiale
- Your City: London
- Your Slack Name:

# Homework Details

- Module: JavaScript
- Week: 5

### DIFF
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -1,17 +1,19 @@
-// Add comments to explain what this function does. You're meant to use Google!
-function getNumber() {
-  return Math.random() * 10;
+// Math.random() can be used to return random integers/numbers
+function getNumber(nim) {
+  return Math.random(nim) * 10;
 }
+console.log(getNumber(23))
 
-// Add comments to explain what this function does. You're meant to use Google!
+// .concat() is used to join two or more strings
 function s(w1, w2) {
-  return w1.concat(w2);
+  return w1.concat(w1, w2);
 }
+console.log (s(34 + ' likes'))
 
 function concatenate(firstWord, secondWord, thirdWord) {
-  // Write the body of this function to concatenate three words together.
-  // Look at the test case below to understand what this function is expected to return.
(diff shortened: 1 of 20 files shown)


mandatory/3-function-output.js
```diff
@@ -1,17 +1,19 @@
...
+function getNumber(nim) {
+  return Math.random(nim) * 10;
 }
+console.log(getNumber(23))
 
-// Add comments to explain what this function does. You're meant to use Google!
+// .concat() is used to join two or more strings
 function s(w1, w2) {
-  return w1.concat(w2);
+  return w1.concat(w1, w2);
```
iTrip23, line 9: I think if you use `w1.concat(w1, w2);` you will get a new string or array that will look like w1w1w2.
let's say you have var w1 = a, and var w2 = b, if you use the syntax like this you will get back as a return 'aab' instead of 'ab'.

mandatory/4-tax.js
```diff
@@ -29,38 +37,3 @@ To run these tests type `node 4-tax.js` into your terminal
 
 ===================================================
 */
-
```
iTrip23, line 32: I don't think we were supposed to delete the test

iTrip23 (commented):
Great job with the homework, there were some minor issues easily fixable.
Keep going!

mandatory/4-tax.js
```diff
@@ -5,7 +5,15 @@
   Sales tax is 20% of the price of the product
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return (price*20/100)
```
akbasmustafa, line 9: There is a little misunderstanding because of function name.  
Comment in exercise: "Remember that the prices must include the sales tax (hint: you already wrote a function for this!)"
So Function must calculate price with tax so It have to be like "return (price*120/100)"
Others look great.

akbasmustafa (changes requested):

PR #77, opened 2021-01-23
London Class 7 - Soraiya Muhammad - JavaScript- Core-1-Week-1

soraiya:
# Your Details

Name: Soraiya Muhammad
City: London
Slack Name: soraiya

# Homework Details

- Module: JS Core 1
- Week: 1


mandatory/4-tax.js
```diff
@@ -17,7 +19,9 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price, percent) {
+  return "£" + price + (0.2 * price).toFixed(2);
+}
```
Jpsi, line 24: The reason this code is returning unexpected values (and hence the tests fail) is that the method `toFixed(2)` is only acting on the expression in brackets, i.e. `(0.2 * price)`.

To fix it, you could put brackets around the whole expression giving you the price with the tax, like so:
`(price + (0.2 * price)).toFixed(2)`.

But what would be even better is to reuse the function `calculateSalesTax` that you have written before, like so:
`calculateSalesTax(price).toFixed(2)`.

Jpsi (commented):
Well done!

PR #82, opened 2021-01-31
South_Africa/JSCore1_Hilton Otto

theTrainMan:
# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

### DIFF
--- a/exercises/B-hello-world/exercise.js
+++ b/exercises/B-hello-world/exercise.js
@@ -1,1 +1,2 @@
 console.log("Hello world");
+console.log('Hello World, I just started learning JavaScript');
--- a/exercises/C-variables/README.md
+++ b/exercises/C-variables/README.md
@@ -12,8 +12,8 @@ The program above will print "Hello world" to the console. Notice how it uses th
 
 ## Exercise
 
-* Add a variable `greeting` to exercise.js (make sure it comes _before_ the console.log)
-* Print your `greeting` to the console 3 times
+- Add a variable `greeting` to exercise.js (make sure it comes _before_ the console.log)
+- Print your `greeting` to the console 3 times
 
 > Remember: to run this exercise you must change directory to the `D-variables`. If you already have a terminal window open for the previous exercise you can do this by running the command `cd ../D-variables`.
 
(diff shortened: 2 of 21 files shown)


RynoFourieZA:
Hi Hilton, I like the way you write your code it is clean and readable.

chandredewet:
Hi Hilton,

Well done as always,I am confident that you are on your way to an excellent programmer and my aim here is simply to highlight certain growth points, for you to look at your work again and become even more. I will put all the points I noticed in this one comment.

F - Exercise 2| The final answer is incorrect as Daniel has 6 characters instead of 9. This has something to do with where in your program you are trimming and calculating the words.... Have a go at it again and if stuck let me know.

I | The answers are also incorrect, as they are not matching the Expected answer given in the README.md file. This is cause I see that you are rounding students and mentors  in line 1 and 2, and then again in 9 and 10. Is this perhaps done in error or were you testing something else. Also something to consider as my previous suggestion for F, where you round the final answer will determine the accuracy of the answer, as you are performing a percentage function too to the same data... So again consider should you round before? or after.

K |  Here the work is all correct, but I saw you did them all in exercise.js rather than in the separate files causing the running of the other files to end up being incorrect. I realize this is the first marking so it was probably for your convenience but as you are moving towards group projects, its important to try and keep to the expected standard getting into a good habit to get into for code readibility and others to know what to expect.

L | When marking this one, I saw what you guys mentioned in the last catchup session, this question was really confusing, and incorrect, and so kudo's to you for attempting to answer it anyway as best you can. The only thing I can point out here though is I still had a look at your first exercise that was commented out, and can see if it was runnable it would not have worked as you not calling function Perc anyway. We can chat about this question in a one on one, but I understand, they did not give clear information or even input variables on this one. But noticed you made sort of a generic function Perc to calculate Percentage but it isn't used at any point. Also as the exercise was to consider nested functions, you should put it in in one of the other functions.

But otherwise, keep up the goodwork, I'm confident your JS will keep getting better.

aashiqabrahams123:
Hi Hilton, well done on your code , clean and readable , Keep up the good work...

PR #90, opened 2021-03-18
South Africa-class 1 AashiqAbrahams JS-core1 week1

aashiqabrahams123:

# Your Details

- Your Name: Aashiq Abrahams
- Your City: South Africa
- Your Slack Name: Aashiq Abrahams

# Homework Details

- Module: JavaScript
- Week:1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Like learning a musical instrument, programming requires daily practise.
 
-The exercises are split into three folders: `exercises`, `mandatory` and `extra`. All homework in the `exercise` and `mandatory` section **must** be completed for homework by the following lesson. 
+The exercises are split into three folders: `exercises`, `mandatory` and `extra`. All homework in the `exercise` and `mandatory` section **must** be completed for homework by the following lesson.
 
 The `extra` folder contains exercises that you can complete to challenge yourself, but are not required for the following lesson.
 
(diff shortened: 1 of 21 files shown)


chandredewet:
Hi Aashiq,,

Well done on completing JS-1- Week 1. And getting through it! Trust the following weeks you will catch up and be able to get even more from class interaction and fellow trainees. Just a few observations from my side...

Exercises:
L| Just to confirm the exercise on CYF side was not clear and in fact erroneous, well done on still trying. I will advise the syllabus team to correct it as they gave two exercises but one file to work on.

Mandatory:
1-syntax-errors.js:
* If you compare expected: 'Hello, my name is Sonjide and I am 27 years old' to what your function returned 'Hello, my name is Sonjideand I am 27years old' there is a small error to fix with the spaces. Correct this until you see the test pass.
* Similarly "return 'The total is ${total}'; " failed because this will only work with backticks, which I see you use in the rest of the homework, so if you can just edit this to pass.

in 3. you added the various variables and strings together without using the concat function. While this will work, try and also learn how to use the concat function for multiple inputs. It would still begin with:
 return firstWord.concat( with all the rest of the input in here);
 
Lastly in 4, inside your function, you make price = calculateSalesTax(price); this is not incorrect but it may be better to rather have had a new variable e.g. priceWithTax, for readability when others read your code. Just a thought here to consider.










K | Here the work is all correct, but I saw you did them all in exercise.js rather than in the separate files causing the running of the o

aashiqabrahams123:
@chandredewet, Thank your for your feedback, I've managed to fix my errors, just a heads up that when I did the commit name is wrong as it should be "Mandatory-Fixed Errors"..

PR #92, opened 2021-05-15
Rewrite tests in Jest

illicitonion:
Fixes https://github.com/CodeYourFuture/syllabus/issues/260

### DIFF
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -1,16 +1,16 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
 function addNumbers(a b c) {
-    return a + b + c;
+  return a + b + c;
 }
 
 function introduceMe(name, age)
-return "Hello, my name is " + name "and I am " age + "years old";
+  return "Hello, my name is " + name "and I am " age + "years old";
 
 function getTotal(a, b) {
-    total = a ++ b;
+  total = a ++ b;
 
-    return "The total is total"
+  return "The total is total";
 }
 
 /* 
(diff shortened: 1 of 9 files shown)


mandatory/1-syntax-errors.js
```diff
@@ -19,24 +19,22 @@ function getTotal(a, b) {
 
 There are some Tests in this file that will help you work out if your code is working.
 
-To run these tests type `node 1-syntax-errors.js` into your terminal
+To run the tests for just this one file, type `npm test -- --testPathPattern 1-syntax-errors` into your terminal
```
textbook, line 22: Isn't `--testPathPattern` effectively the default for positional args? `npm test -- 1-syntax-errors` would work too.

mandatory/1-syntax-errors.js
```diff
@@ -19,24 +19,22 @@ function getTotal(a, b) {
...
+(Reminder: You must have run `npm install` one time before this will work!)
 
 ===================================================
 */
 
-const util = require('util');
+const {test} = require("@jest/globals");
```
textbook, line 28:
```suggestion
const {expect, test} = require("@jest/globals");
```

mandatory/3-function-output.js
```diff
@@ -19,37 +19,21 @@ function concatenate(firstWord, secondWord, thirdWord) {
...
-  "concatenate function - case 2 works",
-  concatenate("I", "like", "pizza"),
-  "I like pizza"
-);
-test(
-  "concatenate function - case 3 works",
-  concatenate("I", "am", 13),
-  "I am 13"
-);
+test("concatenate doesn't only accept strings", () => {
```
textbook, line 37: Good use of test name 👍

mandatory/1-syntax-errors.js
illicitonion: I thought so too, but when I tried it it still ran all of the tests:

```console
% npm test -- 1-currency-conversion

> javascript-core-1-coursework-week1@1.0.0 test
> jest --testRegex='mandatory[/\\].*\.js$' "1-currency-conversion"

 FAIL  mandatory/2-logic-error.js
[...snipped...]
```

mandatory/1-syntax-errors.js
illicitonion: This isn't _required_, but I can see the value of adding it for clarity... When `jest` runs individual test closures, it shoves a bunch of stuff into global scope. But I think it probably is worth explicitly importing, even though we haven't taught the trainees anything about `require` yet and are treating this set-up as a completely opaque box :)

ChrisOwen101 (approved):

PR #104, opened 2021-06-20
 NorthWest Class 4 - Anza Azam - JavaScript - Week 1

Anza-Azam:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/1-currency-conversion.js
+++ b/extra/1-currency-conversion.js
@@ -58,8 +58,8 @@ test("convertToBRL function works for £30", () => {
 
 
 test("convertToBRL function works for £0.99", () => {
-	expect(convertToBRL(1.5)).toEqual(8.46);
-
+  expect(convertToBRL(1.5)).toEqual(8.46);
+});
 test("convertToBRL function works for £1.50", () => {
   expect(convertToBRL(1.5)).toEqual(8.46);
 
--- a/extra/1-currency-conversion.js
+++ b/extra/1-currency-conversion.js
@@ -1,11 +1,17 @@
-/*
-  CURRENCY FORMATTING
-  ===================
-  The business is breaking out into a new market and need to convert prices to USD
-  Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
-*/
+/**
+ * /*
+ *   CURRENCY FORMATTING
+ *   ===================
+ *   The business is breaking out into a new market and need to convert prices to USD
+ *   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
+ *
+ * @format
+ */
 
-function convertToUSD() {}
+function convertToUSD(amount) {
+	var convertedUSD = amount * 1.4;
+	return parseFloat(convertedUSD);
+}
 
 /*
   CURRENCY FORMATTING
@@ -15,7 +21,11 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(price) {
+	var newPrice = (price * 99) / 100;
+	var convertedBRL = (newPrice * 5.7).toFixed(2);
+	return parseFloat(convertedBRL);
+}
 
 /* ======= TESTS - DO NOT MODIFY ===== 
 There are some Tests in this file that will help you work out if your code is working.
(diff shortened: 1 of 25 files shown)


Anza-Azam:
Your Details 
Your Name: Anza Azam
Your City: NorthWest4
Your Slack Name: Anza-Azam
Homework Details
Module: Javascript  Core 1 Module
Week: 1
Notes
What did you find easy?
functions
What did you find hard?
logic
What do you still not understand?
let and var 
Any other notes? Thanks to all polite and patient volunteers

KarenPudner (approved):

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,21 @@
...
 function trimWord(word) {
-  return wordtrim();
+	return word.trim();
 }
 
 function getStringLength(word) {
-  return "word".length();
+
+	var noSpaceString = word.trim();
+	return noSpaceString.length;
```
KarenPudner, line 12: I like the extra work you've done on this -  word.length would have been enough to pass the test, but adding trim is a good idea. When you come to writing tests yourself, think about how you could write a test that would cover this extra functionality.

mandatory/3-function-output.js
```diff
@@ -1,16 +1,25 @@
+/** @format */
+
 // Add comments to explain what this function does. You're meant to use Google!
+
+//This getRandomNumber() gives a random number between 0 and 10
 function getRandomNumber() {
-  return Math.random() * 10;
+	return Math.random() * 10; //Math.random will return a number between 0 and 1
```
KarenPudner, line 7: Is Math.random inclusive of 1? And consequently is your method inclusive of 10?  Check the definition here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random

mandatory/3-function-output.js
```diff
@@ -1,16 +1,25 @@
...
 
 function concatenate(firstWord, secondWord, thirdWord) {
-  // Write the body of this function to concatenate three words together.
-  // Look at the test case below to understand what this function is expected to return.
+	// Write the body of this function to concatenate three words together.
+	// Look at the test case below to understand what this function is expected to return.
+
+	var firstWords = combine2Words(firstWord, " ");
+	var secondWords = combine2Words(secondWord, " ");
+	return firstWords + secondWords + thirdWord;
```
KarenPudner, line 22: This works, but you could do it in one line - .concat can take any number of parameters.

mandatory/4-tax.js
```diff
@@ -1,11 +1,18 @@
...
+ *
+ * @format
+ */
 
-function calculateSalesTax() {}
+//Sales Tax function adds 20% of the price
+function calculateSalesTax(price) {
+	var newPrice = price + (20 / 100) * price;
+	return newPrice;
+}
```
KarenPudner, line 15: This works fine. One change you could make is to declare 20, or 20/100 as a variable called taxRate or something similar, which would make it easier to understand what the code is doing. However, your method has the advantage of being more concise, so it's a matter of taste really, both are correct.

mandatory/3-function-output.js
Anza-Azam: learned now that this will not be inclusive of 1 or 10 so we have to add 1 to make these numbers inclusive, thanks really helped me.

extra/1-currency-conversion.js
```diff
@@ -15,7 +31,11 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(price) {
+	var newPrice = (price * 99) / 100;
+	var convertedBRL = (newPrice * 5.7).toFixed(2);
+	return parseFloat(convertedBRL);
```
KarenPudner, line 37: Maybe assign 5.7 to a variable called something like exchangeRate, just to make the code more understandable

extra/2-piping.js
```diff
@@ -1,41 +1,49 @@
...
-let badCode =
-
+let badCode = add(startingValue, 10);badCode = multiply(badCode, startingValue);badCode = format(badCode);
+//Code is very difficult to read when it is written on single line, makes it challenging to rectify errors if encountered.
 /* BETTER PRACTICE */
 
-let goodCode =
+let goodCode = add(startingValue, 10);
+goodCode = multiply(goodCode, startingValue);
+goodCode = format(goodCode);
```
KarenPudner, line 46: This is much better. I think it could be even easier to understand if you declare a new variable for each line, describing what had been done, i.e. valueAfterAddition. But this is just a personal preference, your way is fine

extra/3-magic-8-ball.js
```diff
@@ -45,10 +93,38 @@
...
+	"Outlook not so good.",
+	"Very doubtful.",
+];
+
 function shakeBall() {
-  //Write your code in here
+	//Write your code in here
+
+	console.log("The ball has shaken!");
+	return answers[Math.floor(Math.random() * 20)];
```
KarenPudner, line 123: You could use answers.length here instead of 20, then you won't need to update the number if you add or remove anything from the array

extra/3-magic-8-ball.js
```diff
@@ -58,61 +134,79 @@ function shakeBall() {
...
+	var status;
+	if (index < 5) {
+		status = "very positive";
+	} else if (index >= 5 && index < 10) {
+		status = "positive";
+	} else if (index >= 10 && index < 15) {
+		status = "negative";
+	} else if (index >= 15 && index < 20) {
+		status = "very negative";
+	}
```
KarenPudner, line 150: Do you need 2 parts to your if statements in lines 144, 146? Have you already established one of them? Do you need the check on line 148 at all?

PR #106, opened 2021-06-21
NW-Class4-Myaskylar-Js-Mandatory-wk1

myaskylar:

# Your Details

- Your Name: Mya Skylar 
- Your City: NW
- Your Slack Name: Mya Skylar

# Homework Details

- Module: JS mandatory
- Week: 1

# Notes

- What did you find easy?
 Functions
- What did you find hard?
 Calculation
 NA
- Any other notes?

### DIFF
--- a/extra/1-currency-conversion.js
+++ b/extra/1-currency-conversion.js
@@ -20,8 +20,9 @@ function convertToUSD(price) {
 
 function convertToBRL(currency) {
   var addFee = currency * 0.01;
-  var foreignRate = (currency - addFee) * 5.7;
-  return parseFloat(foreignRate.toFixed(2));
+  var foreignRate = 5.7;
+  var calculateRate = (currency - addFee) * foreignRate;
+  return parseFloat(calculateRate.toFixed(2));
 }
 
 /* ======= TESTS - DO NOT MODIFY ===== 
--- a/extra/1-currency-conversion.js
+++ b/extra/1-currency-conversion.js
@@ -5,7 +5,10 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(price) {
+  var exchangeRate = price * 1.4;
+  return exchangeRate;
+}
 
 /*
   CURRENCY CONVERSION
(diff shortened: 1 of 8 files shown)


extra/1-currency-conversion.js
```diff
@@ -15,7 +18,11 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(currency) {
+  var addFee = currency * 0.01;
+  var foreignRate = (currency - addFee) * 5.7;
```
KarenPudner, line 23: Very minor point, but I think it would be better to declare 5.7 as the variable foreignRate, and then have a separate variable for the final result, as this makes the code easier to understand

extra/2-piping.js
```diff
@@ -16,26 +16,37 @@
...
+function digit(startingValue) {
+  let calculation = "£" + (startingValue + 10) * 2; return calculation;
+}
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+// This code is difficult to read and look messy.
+let badCode = digit(startingValue);
 
 /* BETTER PRACTICE */
+function goodDigit(startingValue) {
```
KarenPudner, line 44: This is not much different to your badCode. You should write something that calls the 3 functions in one line. So  start off by calling the add function with the startingValue and 2, then wrap that in the multiply function, then wrap that in the format function.

extra/3-magic-8-ball.js
```diff
@@ -59,9 +88,25 @@ function shakeBall() {
   This function should expect to be called with any value which was returned by the shakeBall function.
 */
 function checkAnswer(answer) {
-  //Write your code in here
+  let message = "";
```
KarenPudner, line 91: There is an easier way of doing this using indexOf.  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

extra/3-magic-8-ball.js
```diff
@@ -59,9 +88,25 @@ function shakeBall() {
...
+  for (let i = 0; i < choicesOfAnswer.length; i++) {
+    if (answer === choicesOfAnswer[i]) {
+      message = i;
+    }
+  }
+  if (message < 5) {
+    return "very positive";
+  } else if (message >= 5 && message < 10) {
+    return "positive";
+  } else if (message >= 10 && message < 15) {
```
KarenPudner, line 101: For lines 99 and 101 - do you need 2 parts to your if statement? Have you already established one of them?

extra/1-currency-conversion.js
myaskylar: Thanks Karen. This practice will make the code more understandable. I've made the changes.

extra/3-magic-8-ball.js
myaskylar: Thanks Karen this link makes my code more simple and understand. I've learned more more from here. Thanks for sharing. I've replace my code with using .indexOf and its pass the test! hurray!

PR #110, opened 2021-06-22
NW Class 4 - Maha Malik - JS Core 1 - Week 1

maham511:

# Your Details

- Your Name: Maha Malik
- Your City: NW4
- Your Slack Name: Maha Malik

# Homework Details

- Module: JS Core 1
- Week: 1

# Notes

- What did you find easy?
'Exercises' tasks were fairly easy to understand, but got harder as the tasks progressed. Running tests was very helpful in debugging once I got used to the Jest interface/layout.

- What did you find hard?
Struggle with knowing how and when to store functions into variables, and using variables within functions, it'll probably get clearer with practice. Nesting functions is a bit overwhelming, also knowing where to put parameters in nested functions. 

Same as above

- Any other notes?

### DIFF
--- a/exercises/L-functions-nested/exercise.js
+++ b/exercises/L-functions-nested/exercise.js
@@ -3,3 +3,34 @@ var mentor2 = "Irina";
 var mentor3 = "Mimi";
 var mentor4 = "Rob";
 var mentor5 = "Yohannes";
+
+//THIS SHOULD WORK!
+function createGreeting(name) {
+  var greeting = `Helloooo ${name}`;
+  return greeting.toUpperCase();
+}
+
+console.log(createGreeting(mentor1));
+console.log(createGreeting(mentor2));
+console.log(createGreeting(mentor3));
+console.log(createGreeting(mentor4));
+// console.log(createGreeting(mentor5));
+
+//UPTO HERE
+
+// // function CreateShoutyGreeting(){
+// //     return createGreeting.toUpperCase();
+// // }
+
+// CREATES ALL CAPS FOR A NAME and logs
+// function createCapName(name){
+//     return name.toUpperCase();
+// }
+// // console.log(createCapName(mentor4));
+
+// function createShoutyGreeting(name){
+//     var message = createCapName(name);
+//     return `Hello ${name}`;
+// }
+// console.log(createShoutyGreeting(mentor4));
+
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -10,9 +10,10 @@ function combine2Words(word1, word2) {
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
-  // Write the body of this function to concatenate three words together.
-  return firstWord + " " + secondWord + " " + thirdWord;  
+  return `${firstWord} ${secondWord} ${thirdWord}`; 
 }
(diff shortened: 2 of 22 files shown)


mandatory/3-function-output.js
```diff
@@ -1,17 +1,19 @@
-// Add comments to explain what this function does. You're meant to use Google!
+//My Comment - This function gives a random number by using math.random() and multiplying the value by 10
```
KarenPudner, line 1: Can you explain the range of the random numbers that will be returned? What range does Math.random() return, and so what will your method return?

mandatory/4-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(num) {
+  return num *1.2;    //Adds 20% to number and gives total price including tax
```
KarenPudner, line 9: You could also declare 1.2 as a variable called priceWithTax or something similar, or even declare 0.2 as a variable call tax, and then multiply by the price and then add to the price. This would would make it easier to understand what your code is doing. However, your solution has the advantage of being concise. Both approaches work, so it's a matter of taste.

mandatory/4-tax.js
```diff
@@ -15,9 +17,18 @@ function calculateSalesTax() {}
...
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(num) {
+   return `£${calculateSalesTax(num).toFixed(2)}`;
+}
+
+//CAN ALSO WRITE THIS WAY:
+// function addTaxAndFormatCurrency(num) {
+//   var totalPrice= `£${calculateSalesTax(num).toFixed(2)}`;
```
KarenPudner, line 29: Yes, as discussed in the previous comment, the first is more concise, but sometimes breaking your code down into stages and declaring variables with good names helps to make your code more understandable. It's up to you which you prefer.

exercises/L-functions-nested/exercise.js
```diff
@@ -3,3 +3,34 @@ var mentor2 = "Irina";
 var mentor3 = "Mimi";
 var mentor4 = "Rob";
 var mentor5 = "Yohannes";
+
+//THIS SHOULD WORK!
+function createGreeting(name) {
+  var greeting = `Helloooo ${name}`;
+  return greeting.toUpperCase();
+}
```
lannem, line 11: the idea of this exercise was to have 2 separate functions, one that returns the name in uppercase and another that returns the greeting by calling the first function to get the name

mandatory/3-function-output.js
```diff
@@ -1,17 +1,19 @@
...
+
+//My comment - The function combines two words using the concat method.
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
-  // Write the body of this function to concatenate three words together.
-  // Look at the test case below to understand what this function is expected to return.
+  return `${firstWord} ${secondWord} ${thirdWord}`;
```
lannem, line 13: can you do this using the `.concat()` method?

lannem (commented):
Nice work Maha :)

PR #112, opened 2021-06-22
NW4 - Sharmaine Taylor - JS - Week1

Sharm-Dev:

# Your Details

- Your Name: Sharmaine Taylor
- Your City: NW
- Your Slack Name: Sharmaine

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/E-strings-concatenation/exercise.js
+++ b/exercises/E-strings-concatenation/exercise.js
@@ -1,3 +1,5 @@
 // Start by creating a variable `message`
-
-console.log(message);
+let greetingStart = 'Hello, my name is '
+let name = 'Sharm'
+let greeting = greetingStart + name;
+console.log(greeting);
--- a/exercises/F-strings-methods/exercise.js
+++ b/exercises/F-strings-methods/exercise.js
@@ -1,3 +1,8 @@
 // Start by creating a variable `message`
-
+let name = 'Sharm';
+let nameLength = name.length;
+const message = 'My name is ' + `${name}` + ' and my name is ' + `${nameLength}` + ' characters long'
 console.log(message);
+
+
+
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,3 +1,4 @@
-const name = " Daniel  ";
-
-console.log(message);
+const newName = " Daniel  ".trim();
+const newNameLength = newName.length;
+const newMessage = 'My name is ' + `${newName}` + ' and my name is ' + `${newNameLength}` + ' charactors long'
+console.log(newMessage);
--- a/exercises/J-functions/exercise.js
+++ b/exercises/J-functions/exercise.js
@@ -1,7 +1,10 @@
 function halve(number) {
   // complete the function here
+  return number / 2;
 }
 
 var result = halve(12);
+var results = halve(20);
 
 console.log(result);
+console.log(results);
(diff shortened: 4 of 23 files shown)


exercises/E-strings-concatenation/exercise.js
```diff
@@ -1,3 +1,5 @@
 // Start by creating a variable `message`
-
-console.log(message);
+let greetingStart = 'Hello, my name is '
+let name = 'Sharm'
```
MCompton96, line 3: It's a small thing here, but try to get into the habit of assigning variables that you're not planning to change as const instead of let, you'll understand later when you get onto scope but feel free to message me if you don't understand

exercises/F-strings-methods/exercise.js
```diff
@@ -1,3 +1,8 @@
 // Start by creating a variable `message`
-
+let name = 'Sharm';
+let nameLength = name.length;
+const message = 'My name is ' + `${name}` + ' and my name is ' + `${nameLength}` + ' characters long'
```
MCompton96, line 4: Unfortunately not the correct implementation of string literal syntax here, try playing around with having more than one variable in a string and you might see the usefulness of them...

Let me know if you're still unsure

exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,4 @@
-const name = " Daniel  ";
-
-console.log(message);
+const newName = " Daniel  ".trim();
```
MCompton96, line 1: You don't need to delete the original name variable, you can call the trim method on that variable and you get the same result, this would be a cleaner implementation

exercises/J-functions/exercise.js
```diff
@@ -1,7 +1,10 @@
 function halve(number) {
   // complete the function here
+  return number / 2;
 }
 
 var result = halve(12);
+var results = halve(20);
```
MCompton96, line 7: Be careful when naming variables very similarly, also try not to use var, use const or let if you can var is a bit more old school

extra/1-currency-conversion.js
```diff
@@ -15,7 +19,12 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(pound) {
+  let fee = 99/100;
+  let brazilianPrice = pound * fee;
+  let exchangeRate = brazilianPrice * 5.7;
+  return Number((exchangeRate.toFixed(2)));
```
MCompton96, line 26: Good solution Sharmaine!!

mandatory/4-tax.js
```diff
@@ -5,7 +5,11 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  let tax = price * 20/100;
+  let newPrice = price + tax;
+  return newPrice; 
+}
```
MCompton96, line 12: Good!

mandatory/4-tax.js
```diff
@@ -17,7 +21,11 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price) {
+  let priceAddTax =calculateSalesTax(price);
```
MCompton96, line 25: Really nice to see you using a function you've already created inside another function!

PR #119, opened 2021-06-23
NW4 - Azan Shaikh - JS - Week 1  

Ash-97:

# Your Details

- Your Name: Azan Shaikh
- Your City: NorthWest4
- Your Slack Name: Azan

# Homework Details

- Module: JavaScript
- Week: 1

# Notes

- What did you find easy?
   Exercises
- What did you find hard?
   Reading Jest results.
   
- Any other notes?

### DIFF
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,6 +1,6 @@
 const name = " Azan  ";
-let nametrim = name.trim();
-let nameLength = name.length;
+let nameTrim = name.trim();
+let nameLength = nameTrim.length;
 
 console.log(
   "My name is",
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,3 +1,11 @@
-const name = " Daniel  ";
+const name = " Azan  ";
+let nametrim = name.trim();
+let nameLength = name.length;
 
-console.log(message);
+console.log(
+  "My name is",
+  nametrim,
+  "and it is",
+  nameLength,
+  "characters long."
+);
--- a/exercises/L-functions-nested/exercise.js
+++ b/exercises/L-functions-nested/exercise.js
@@ -1,5 +1,29 @@
-var mentor1 = "Daniel";
-var mentor2 = "Irina";
-var mentor3 = "Mimi";
-var mentor4 = "Rob";
-var mentor5 = "Yohannes";
+const mentor1 = "Daniel";
+const mentor2 = "Irina";
+const mentor3 = "Mimi";
+const mentor4 = "Rob";
+const mentor5 = "Yohannes";
+
+function shoutyGreeting(mentors) {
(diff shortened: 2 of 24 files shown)


exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,11 @@
-const name = " Daniel  ";
+const name = " Azan  ";
+let nametrim = name.trim();
+let nameLength = name.length;
```
KarenPudner, line 3: You should get the length of nameTrim rather than name

exercises/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,29 @@
...
+const shoutyGreeting3 = createGreeting(mentor3);
+const shoutyGreeting4 = createGreeting(mentor4);
+const shoutyGreeting5 = createGreeting(mentor5);
+
+console.log(shoutyGreeting1);
+console.log(shoutyGreeting2);
+console.log(shoutyGreeting3);
+console.log(shoutyGreeting4);
+console.log(shoutyGreeting5);
+*/
```
KarenPudner, line 29: It is good to get into the habit of deleting any experimental code that you have commented out.

mandatory/3-function-output.js
```diff
@@ -1,16 +1,19 @@
 // Add comments to explain what this function does. You're meant to use Google!
+/* Math.random function returns a pseudo random value between 0 to 1 and multiplies it by 10.*/
```
KarenPudner, line 2: Is the range inclusive of 1? And consequently, will the range of output from your method be inclusive of 10?

mandatory/3-function-output.js
```diff
@@ -1,16 +1,19 @@
...
 function getRandomNumber() {
   return Math.random() * 10;
 }
 
 // Add comments to explain what this function does. You're meant to use Google!
+/* Concat method is used to merge two or more arrays and it does not change the existing array but returns a new array.*/
```
KarenPudner, line 8: It can be used as a method on an array, but here it is being used on a string. It behaves in the same way on both types.

exercises/F-strings-methods/exercise2.js
Ash-97: Thank you for the feedback, I will be sure to fix it.

exercises/L-functions-nested/exercise.js
Ash-97: I was thinking about deleting it so it doesn't look cluttered but decided to leave it so I would remember the different ways that the exercise could solved.

mandatory/3-function-output.js
Ash-97: Added a bit more detail to the comment, Thanks Karen.

mandatory/3-function-output.js
Ash-97: Changed the comment to mention strings as well as arrays. :)

PR #129, opened 2021-06-25
North West 4 - Morteza Khojasteh - JavaScript - Week 1 (Mandatory)

Morteza-Khojasteh:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -5,12 +5,15 @@ function getRandomNumber() {
 
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
-  return word1.concat(word2);
+  const firstWord = word1.concat(" ", word2);
+  return firstWord ;
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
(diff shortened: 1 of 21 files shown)


mandatory/3-function-output.js
```diff
@@ -5,12 +5,15 @@ function getRandomNumber() {
```
lannem, line 5: can you explain what the first 2 functions do?

lannem (commented):
Nice work :)

Morteza-Khojasteh:
> Nice work :)

Thanks

mandatory/3-function-output.js
Morteza-Khojasteh: I have a function **getRandomNumber()** without parameter inside the function there is another function The **Math. random() function** returns a floating-point, pseudo-random number in the range 0 to less than 1 (inclusive of 0, but not 1) with approximately uniform distribution over that range — which you can then scale to your desired range
// expected output: a number from 0 to <10

PR #176, opened 2021-07-10
GLASGOW_CLASS_5_OMER_ALI_WEEK_1

Omer249a:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/F-strings-methods/exercise.js
+++ b/exercises/F-strings-methods/exercise.js
@@ -1,3 +1,13 @@
 // Start by creating a variable `message`
+const MY_NAME = "Omer";
+const NAME_LENGTH = MY_NAME.length;
+let message = "";
 
+// Update Message & output message
+message =
+  "My name is " +
+  MY_NAME +
+  " and my name is " +
+  NAME_LENGTH +
+  " characters long";
 console.log(message);
--- a/exercises/G-numbers/exercise.js
+++ b/exercises/G-numbers/exercise.js
@@ -1,1 +1,14 @@
 // Start by creating a variables `numberOfStudents` and `numberOfMentors`
+let numberOfStudents = 15;
+let numberOfMentors = 8;
+let totalStudentsAndMentors = 0;
+let message = "";
+
+// Calculate total number of students
+totalStudentsAndMentors = numberOfStudents + numberOfMentors;
+
+// Concatenate string
+message = `Number of students: ${numberOfStudents}
+Number of mentors: ${numberOfMentors}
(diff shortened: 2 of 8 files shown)


exercises/F-strings-methods/exercise.js
```diff
@@ -1,3 +1,13 @@
...
+let message = "";
 
+// Update Message & output message
+message =
+  "My name is " +
+  MY_NAME +
+  " and my name is " +
+  NAME_LENGTH +
+  " characters long";
 console.log(message);
```
alastair87, line 13: Excellent.

exercises/G-numbers/exercise.js
```diff
@@ -1 +1,14 @@
 // Start by creating a variables `numberOfStudents` and `numberOfMentors`
+let numberOfStudents = 15;
+let numberOfMentors = 8;
+let totalStudentsAndMentors = 0;
+let message = "";
+
+// Calculate total number of students
+totalStudentsAndMentors = numberOfStudents + numberOfMentors;
```
alastair87, line 8: You can simplify this by initialising the variable and setting its value at the same time.

I.e.:

`let totalStudentsAndMentors = numberOfStudents + numberOfMentors;`

(so you no longer need line 4)

exercises/G-numbers/exercise.js
```diff
@@ -1 +1,14 @@
...
+
+// Concatenate string
+message = `Number of students: ${numberOfStudents}
+Number of mentors: ${numberOfMentors}
+Total number of students and mentors: ${totalStudentsAndMentors}`;
+console.log(message);
```
alastair87, line 14: Excellent. Again, you could do `let message = ` on line 11 and define the variable at the same time as you set its value. This can sometimes be safer, as it means you'll get an error message saying the variable is `undefined` if you try to use it before it has been set correctly. This is helpful as then you'll know there's something wrong.

exercises/J-functions/exercise.js
```diff
@@ -1,5 +1,6 @@
 function halve(number) {
   // complete the function here
+  return (result = number / 2);
 }
 
 var result = halve(12);
```
alastair87, line 6: Excellent.

alastair87 (commented):
Good  work on parts A through G, though you should try and have a look at I through L. I can arrange a time to look at some of these with you if that would help.

PR #177, opened 2021-07-14
Add GitHub Actions to run tests

illicitonion:
There are two separate actions - one for the mandatory tests, and one
for the extra tests which is only run if any extra tests were modified.

There's also a fork of someone else's GitHub Jest reporter in here to
post in-line comments with test failures. Unfortunately, this has been
pending merge into Jest for quite a while, so if we want to use it we
need to fork it.

It could be nice, rather than forking it into each of our weeks of
coursework, to release the code as a package and depend on it, but I'm
not sure about the license story of doing so.

This currently juset displays a status, but we could also make it post
results somewhere if we had somewhere to aggregate results.

Note: This in no way checks that tests haven't been deleted or modified in sketchy ways.

Example PRs:
* [Failed some mandatory exercises](https://github.com/illicitonion/JavaScript-Core-1-Coursework-Week1/pull/2)
* [Completed all mandatory exercises but didn't try any extras](https://github.com/illicitonion/JavaScript-Core-1-Coursework-Week1/pull/3)
* [Completed all mandatory exercises, tried some extra ones but didn't get them all right](https://github.com/illicitonion/JavaScript-Core-1-Coursework-Week1/pull/4)
* [Completed all mandatory exercises, tried or deleted all extra ones](https://github.com/illicitonion/JavaScript-Core-1-Coursework-Week1/pull/5)


.github/workflows/mandatory-tests.yml
```diff
@@ -0,0 +1,11 @@
...
+on: pull_request
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install modules
+      run: npm install
+    - name: Run mandatory tests
+      run: npm test
```
gregdyke, line 11: I think it would be good if npm test runs all tests, so it would just be a different command for the mandatory tests (will need to remember to change this when making that change)

PR #178, opened 2021-07-21
NW4-Haliema Adiem-Javascript-core1-coursework-week1

Haliema26:

# Your Details

- Your Name: Haliema Adiem
- Your City: Northwest class
- Your Slack Name: Haliema

# Homework Details

- Module: Javascript
- Week: 1

# Notes

- What did you find easy?
- some parts of the  exercise task

- What did you find hard?
- The Mandatory task

- still confused about the different styles of writing functions.
- find it hard sometimes to draw a plan for my answer before starting answering the questions directly.  

- Any other notes?

### DIFF
--- a/exercises/B-hello-world/exercise.js
+++ b/exercises/B-hello-world/exercise.js
@@ -1,1 +1,7 @@
-console.log("Hello world");
+console.log("Hello World. I just started learning JavaScript!");
+//Q What happens when you get rid of the quote marks?
+//A it does not run and it gives syntax error.
+//Q What happens when you console.log() just a number without quotes?
+//A it shows the number in terminal window after running the command node.
+console.log(5);
+
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,3 +1,6 @@
 // Start by creating a variable `greeting`
+var greeting = "Hello world";
 
+console.log(greeting);
+console.log(greeting);
 console.log(greeting);
--- a/exercises/D-strings/exercise.js
+++ b/exercises/D-strings/exercise.js
@@ -1,3 +1,7 @@
 // Start by creating a variable `message`
 
+var message = "This is a string";
+var messageType = typeof message;
+
 console.log(message);
+console.log(messageType);
--- a/exercises/E-strings-concatenation/exercise.js
+++ b/exercises/E-strings-concatenation/exercise.js
@@ -1,3 +1,8 @@
 // Start by creating a variable `message`
+var greetingStart = "Hello, my name is ";
+var name = "Haliema";
 
-console.log(message);
+var greeting = greetingStart + name;
+
+console.log(greeting); // Logs "Hello, my name is Haliema"
+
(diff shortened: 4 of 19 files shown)


SteveLeicester:
Great to see this Haliema.  All the exercises seem fine and work apart from L-functions-nested. Well actually your L-function-nested does give the expected output.   But take care to follow the instructions.

So we are asked to **include a function that spells the name in uppercase**

Lets do this - and call it shoutyName 

`function shoutyName(person){`
`        return person.toUpperCase();`
`}`

We are also asked to make a shouty greeting which will be "HELLO", So we get

`function shoutyGreeting(person){`
`    var greeting = "HELLO ";`

`    /* Now we can call the shoutyName function from   within this function.  */ `
`    return greeting + shoutyName(person);`
`}`

Then run the functions

`console.log(shoutyGreeting(mentor1));`
`console.log(shoutyGreeting(mentor2));`
`console.log(shoutyGreeting(mentor3));`
`console.log(shoutyGreeting(mentor4));`
`console.log(shoutyGreeting(mentor5));`




Can you think of an advantage of doing this compared to your way? Both give the same result so your customer will be happy with the software you provided.

What if the customer said can you change the greeting to "HELLO, HOW NICE TO SEE YOU ".  How many lines of code will you have to change compared to the above nested function approach.

Perhaps your customer wont mind but if there are thousands of lines of code and a change needs to be made it can make a huge difference to how long the customer will get the new software who may no longer be happy to hear that.

But the key take away is to follow the instructions. Oh and to ask if the instructions don't make sense.

Now - on to the mandatory work.  Now I know you said you found these hard but great that you had a go.  Did you manage to run the tests using jest?

Be careful when concatenating strings 

`var string1 = "hello"`
`var string2 = "Haliema"`
`var string3 = string1 + string2  /*Leaves no space so check whether you need to ensure there is a gap between strings */`


The test fails as follows

Expected: "Hello, my name is Sonjide and I am 27 years old"
Received: "Hello, my name is Sonjide and I am 27years old"

Its only a minor thing but just to say be careful when joining strings.

The other thing which is important - TTD. Test Driven Development.   The idea is that a test tells you what the function should do. Its like the instructions and so important to look at the tests then write a function to pass the test.

`test("getTotal returns a string describing the total", () => {`
`  expect(getTotal(23, 5)).toEqual("The total is 28");`
`});`

So getTotal returns a string describing the total - this is very important.

So lets make the function to do just that.

`function getTotal(a, b){`
`    var sum = a + b;`
`    /* We have the sum - so make a description */`
`    return "The total is " + sum;`
`}`

So if you look at the tests - they are telling you what to do

Note there are similar space problems with concat function in 3-function-output.  So yes care with space between concatenated strings. See what the expected result is "code_space_your_space_future".

`test("concatenate example #1", () => {`
`  expect(concatenate("code", "your", "future")).toEqual("code your future");`
`});`

Hopefully looking at the tests will help to make the tasks more straight forward.

But this is a good effort and I think the most important thing is to take care with the instructions, study the tests for TTD and make a function to pass those tests (for the mandatory exercises).

Keep it going Haliema.

Steve

Haliema26:
Thanks, @SteveLeicester, I can see more broadly now. I have to put on my account the simplicity, and it is implemented in the fewer lines of coding. similarly, I have to consider getting back to my code later either for debugging or adding new features and then this is the job of the developer.

PR #221, opened 2021-12-11
LONDON CLASS 8 - BERKELI HALMYRADOV - JavaScript - WEEK1

berkeli:

# Your Details

- Your Name: Berkeli Halmyradov
- Your City: London
- Your Slack Name: Berkeli

# Homework Details

- Module: JavaScript
- Week: 1

# Notes

- What did you find easy?
- Variables, functions

- What did you find hard?
- Testing was a bit difficult and I ran into a problem where one of the test functions has been deprecated from Jest (toBeOneOf), I couldn't get the test to pass because of  this error but I added another test which does the same thing.

- All clear

- Any other notes?
- no

@gixo This week's homework. Thank you :)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,20 @@
+module.exports = {
+  env: {
+    browser: true,
+    es2021: true,
+    node: true,
+    es6: true,
+    'jest/globals': true,
+  },
+  extends: ['airbnb-base', 'plugin:jest/recommended'],
+  parserOptions: {
+    ecmaVersion: 13,
+  },
+  rules: {
+    'max-len': 0,
+    'no-unused-vars': 'warn',
+    quotes: ['error', 'double', { avoidEscape: true }],
+    'no-console': 'off',
+  },
+  plugins: ['jest', 'prettier'],
+};
--- a/exercises/K-functions-parameters/exercise.js
+++ b/exercises/K-functions-parameters/exercise.js
@@ -1,9 +1,11 @@
 // Complete the function so that it takes input parameters
-function multiply() {
-  // Calculate the result of the function and return it
+function multiply(a, b) {
+  return a * b;
 }
 
 // Assign the result of calling the function the variable `result`
 var result = multiply(3, 4);
 
 console.log(result);
+console.log(multiply(2, 1));
+console.log(multiply(112, 34));
--- a/exercises/K-functions-parameters/exercise4.js
+++ b/exercises/K-functions-parameters/exercise4.js
@@ -1,5 +1,5 @@
 const addNumbers = (a, b) => a + b;
 
-let sum = addNumbers(13, 124)
+const sum = addNumbers(13, 124)
 
 console.log(sum);
--- a/exercises/K-functions-parameters/exercise4.js
+++ b/exercises/K-functions-parameters/exercise4.js
@@ -1,5 +1,5 @@
-// Declare your function first
+const addNumbers = (a, b) => a + b;
 
-// Call the function and assign to a variable `sum`
+let sum = addNumbers(13, 124)
 
 console.log(sum);
--- a/extra/1-currency-conversion.js
+++ b/extra/1-currency-conversion.js
@@ -5,7 +5,9 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(amount) {
+  return Math.floor((amount * 1.4) * 100) / 100;
+}
 
 /*
   CURRENCY CONVERSION
@@ -15,7 +17,9 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(amount) {
+  return Math.floor((amount * 0.99 * 5.7) * 100) / 100;
+}
 
 /* ======= TESTS - DO NOT MODIFY ===== 
 There are some Tests in this file that will help you work out if your code is working.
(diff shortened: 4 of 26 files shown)


gixo:
Thanks for submitting @berkeli! Looks like not all tests are passing though, can you look into this before I start my review?

berkeli:
> Thanks for submitting @berkeli! Looks like not all tests are passing though, can you look into this before I start my review?

Hi Alebrto, I have fixed it now. 
There was an issue with `jest-extended` not being recognised correctly, I could only resolve it by importing the module manually.

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,17 @@
...
+function introduceMe(name, age) {
+  return "Hello, my name is " + name + " and I am " + age + " years old";
+}
+  
 
 function getTotal(a, b) {
-  total = a ++ b;
-
-  return "The total is total";
+  total = a + b;
```
gixo, line 13: This variable has been initialized without it being defined first.
```suggestion
  const total = a + b;
```

gixo (changes requested):
Well done Berkeli, as usual this is a great submission!

My main remarks are the following: 
 - remember to always declare variables before using them
 - ensure you use `let` and `const` appropriately
 - in general, prefer the use of template literals over manual concatenation
 - use an automatic linter (as explained below)

Many of these problems would have been caught by a linter enforcing formatting rules. I suggest you install `eslint `with the Airbnb style guide to see how it works and to study the suggestions that are provided. You can also set up VS Code to automatically fix your code according to the eslint rules.

See this article about installing eslint: https://eslint.org/docs/user-guide/getting-started 

When initializing eslint (`eslint --init`) select the third option (`To check syntax, find problems, and enforce code style`) and then select `Airbnb` for code style presets. You can then install the [jest plugin](https://www.npmjs.com/package/eslint-plugin-jest) for eslint and use the following settings in your `.eslintrc.js` config file.

```
module.exports = {
  env: {
    browser: true,
    es2021: true,
    "jest/globals": true,
  },
  extends: [
    "airbnb-base",
    "plugin:jest/recommended",
  ],
  parserOptions: {
    ecmaVersion: 13,
  },
  rules: {
    "max-len": 0,
    quotes: ["error", "double", { avoidEscape: true }],
    "no-console": "off",
  },
  plugins: ["jest"],
};

```

I can have another look at the code once you have implemented these changes.

exercises/K-functions-parameters/exercise4.js
```diff
@@ -1,5 +1,5 @@
-// Declare your function first
+const addNumbers = (a, b) => a + b;
 
-// Call the function and assign to a variable `sum`
+let sum = addNumbers(13, 124)
```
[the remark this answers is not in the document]
berkeli: Thanks Alberto, I have fixed these issues now. I have ESlint installed, but couldn't get it to work (doesn't highlight errors), I will do some further testing and try to get it to work.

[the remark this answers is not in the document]
gixo: eslint might be complicated to install the first time. Have you installed it globally perhaps? That is not recommended, as its plugins also need to be installed globally if you do so. See installation guide here: https://eslint.org/docs/6.0.0/user-guide/getting-started

You can troubleshoot eslint errors by installing the [eslint vs code extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint), and then checking the eslint output log in vs code. For example, here is what happens if I remove the eslint jest plugin from the .eslintrc.js file:

image

.eslintrc.js
```diff
@@ -0,0 +1,20 @@
...
+  parserOptions: {
+    ecmaVersion: 13,
+  },
+  rules: {
+    'max-len': 0,
+    'no-unused-vars': 'warn',
+    quotes: ['error', 'double', { avoidEscape: true }],
+    'no-console': 'off',
+  },
+  plugins: ['jest', 'prettier'],
```
gixo, line 19: Looks like you are now using prettier - I would try to stick with eslint only for this project (disable prettier for the vs code workspace). eslint and prettier can conflict with each other.

exercises/K-functions-parameters/exercise.js
```diff
@@ -1,9 +1,11 @@
 // Complete the function so that it takes input parameters
-function multiply() {
-  // Calculate the result of the function and return it
+function multiply(a, b) {
+  return a * b;
```
Selchuk-Karakus, line 3: Avoid using single letters.  It can be difficult to understand the purpose of the variable. Best to use clear variable names.

Selchuk-Karakus (commented):
I would make the proposed changes to the code to make the code be easier to read and understand.

extra/1-currency-conversion.js
```diff
@@ -15,7 +17,9 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(amount) {
+  return Math.floor((amount * 0.99 * 5.7) * 100) / 100;
```
gixo, line 21: Here, for better readability, you should separate this computation into separate variables, for example:
 - What is the transaction fee?
 - What's the exchange rate?
 - what's the amount after the fee?

It may look like too much for a simple computation, but the idea with these exercises is to start practicing your code readability skills that will be very useful in the future :)

gixo (changes requested):
Added one remark about code readability, also see [this comment](https://github.com/CodeYourFuture/JavaScript-Core-1-Coursework-Week1/pull/221#discussion_r767356219) for how to troubleshoot eslint setup.

PR #232, opened 2021-12-13
LONDON_8-SELCHUK_KARAKUS-JS1-WEEK_1

Selchuk-Karakus:

# Your Details

- Your Name: Selchuk Karakus
- Your City: London
- Your Slack Name: Selchuk Karakus

# Homework Details

- Module: JS
- Week: 1

# Notes

- What did you find easy?
It's all quite easy for me at this stage.

- What did you find hard?
Magic 8 ball challenge. Getting the code to pass the Jest test was impossible. Got help from Berkeli.

It's all very clear so far.

- Any other notes?

### DIFF
--- a/exercises/E-strings-concatenation/exercise.js
+++ b/exercises/E-strings-concatenation/exercise.js
@@ -1,3 +1,12 @@
-// Start by creating a variable `message`
-
+// Start by creating a variable `message` //
+let message = "Hello, my name is ";
 console.log(message);
+
+// - Write a program that logs a message with a greeting and your name //
+const greetPerson = (person) => {
+  return message + person;
+}
+
+let personOne = "Daniel";
+let result = greetPerson(personOne);
+console.log(result);
--- a/extra/1-currency-conversion.js
+++ b/extra/1-currency-conversion.js
@@ -5,7 +5,9 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(amount) {
+  return amount * 1.4;
+}
 
 /*
   CURRENCY CONVERSION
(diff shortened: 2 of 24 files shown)


exercises/E-strings-concatenation/exercise.js
```diff
@@ -1,3 +1,12 @@
-// Start by creating a variable `message`
-
+// Start by creating a variable `message` //
+let message = "Hello, my name is ";
```
berkeli, line 2:
```suggestion
const message = "Hello, my name is ";
```
It's advisable to use **const** if we are not modifying the value

extra/1-currency-conversion.js
```diff
@@ -15,7 +17,13 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+const convertToBRL = (amount) => {
+  let deductionFee = amount / 100;
+  let poundToConvert = amount - deductionFee;
+  let payOut = +(poundToConvert * 5.7).toFixed(2);
```
berkeli, line 23: toFixed converts a number into a string. rather than converting it to string, and then back to number with the plus sign, we can use Math.floor(x * 100) / 100 to round the number to 2 decimal places.

berkeli (commented):
Great Work Selchuk! 

No issues as far as I can see, just a note on using const instead of let if we are not changing the variable.

berkeli:
One more thing 
I noticed your tests are failing on GitHub (the red cross next to pull request). This is because tests are not correcyl defined in package.json file. 

you can add the following code in scripts section, which should fix the issue

` "test": "jest",
    "extra-tests": "jest extra"
`

Selchuk-Karakus:
> "test": "jest", "extra-tests": "jest extra"

Nice one!

PR #240, opened 2021-12-15
London-class-8-Helena-Rog-JS-Week1

hachi-ops:

# Your Details

- Your Name: Helena Rog
- Your City: London
- Your Slack Name: Helena Rog

# Homework Details

- Module: JS
- Week: 1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


exercises/C-variables/exercise.js
```diff
@@ -1,3 +1,7 @@
 // Start by creating a variable `greeting`
 
+var greeting = "Hello World!"
```
s-shah-coder, line 3: As per CYF guidelines we should refrain from using var when declaring a variable and instead use let and const.

hachi-ops: I've been guilty of this mistake all throughout my code:)

exercises/F-strings-methods/exercise.js
```diff
@@ -1,3 +1,13 @@
 // Start by creating a variable `message`
+var myName = "Helen";
+var myNameLength = myName.length;
+var message = "Hello, my name is " + myName + "and my name is " + myNameLength + " characters long."
```
jstadnik, line 4: This could be a good place to use string interpolation, i.e
```suggestion
var message = `Hello, my name is ${myName} and my name is ${myNameLength} characters long.`
```

exercises/K-functions-parameters/exercise4.js
```diff
@@ -1,5 +1,11 @@
 // Declare your function first
 
+function sum (num1, num2) {
+    return num1 + num2;
+}
+    var sum = sum(13, 124);
```
jstadnik, line 6: Indentation :)

exercises/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,29 @@
-var mentor1 = "Daniel";
-var mentor2 = "Irina";
-var mentor3 = "Mimi";
-var mentor4 = "Rob";
-var mentor5 = "Yohannes";
+    var mentor1 = "Daniel";
```
jstadnik, line 1: There's some indentation issues in the file -- basically, everything should start from the very left side, unless it's inside function brackets, or a for loop (you'll learn about these later) or something of this sort.

exercises/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,29 @@
...
+    var mentor2 = "Irina";
+    var mentor3 = "Mimi";
+    var mentor4 = "Rob";
+    var mentor5 = "Yohannes";
+
+function helloGreeting(name) {
+    return name;     
+}
```
jstadnik, line 9: This function doesn't do anything as is -- all the work is being done by the other function. As per instructions, you should have:
* One function that returns the name in all caps
* One function that prints "HELLO <NAME>"
Note, the second function should call your first function to achieve its goal. See how it's used for age in the example in the readme for this exercise.

exercises/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,29 @@
...
+
+function helloGreeting(name) {
+    return name;     
+}       
+    function upperCaseGreeting(helloGreeting) {
+        return ("Hello " + helloGreeting).toUpperCase();   
+}
+
+var greeting = helloGreeting;
+var shoutyGreeting = upperCaseGreeting;
```
jstadnik, line 15: This is just renaming the function -- is this what you intended to do?

jstadnik (commented):
Great work! Only issue I've noticed is that you should try working on reusing functions.

Basically if I already have a function
```
function add(a, b) {
   return a + b;
}
```
And I want to create a function that returns string "The sum of 2 and 3 is 5" I would do this by writing
```
function printSum(a, b) {
     let sum = add(a, b);
     return `The sum of ${a} and ${b} is ${sum}`;
}
```
You can see I'm using the "add" function here, rather than write "a+b" again. This may seem silly for such basic examples, but it's the DRY (don't repeat yourself) rule that matters more once your functions do more complicated stuff.

The functions you write, can be used exactly like the built-in JavaScript functions (e.g. `Math.round`).

PR #244, opened 2021-12-15
London Class_8 - Deago Browne - JS - Week_1

DeagoBrowne:

# Your Details

- Your Name: Deago Browne
- Your City: London
- Your Slack Name: Deago Browne - Trainee

# Homework Details

- Module: JS1

# Notes

- What did you find easy? **Most of the early exercises, and most of the npm Tests**

- What did you find hard? **Percentages - though I managed to get it done, and I also found some formulas to use in the future**

- What do you still not understand? **Couldn't add the "£" to the Tax function - gave up, as it was time-consuming, and I have other assignments to complete**

- Any other notes?

-----
[View rendered exercises/B-hello-world/README.md](https://github.com/DeagoBrowne/JavaScript-Core-1-Coursework-Week1/blob/master/exercises/B-hello-world/README.md)

### DIFF
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -1,16 +1,19 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a, b, c) {
   return a + b + c;
 }
 
-function introduceMe(name, age)
-  return "Hello, my name is " + name "and I am " age + "years old";
+var name = "Sonjide";
+
+function introduceMe(name, age) {
+  return "Hello, my name is " + name + " and I am " + age + " years old";
+}
 
 function getTotal(a, b) {
-  total = a ++ b;
+  total = a + b;
 
-  return "The total is total";
+  return "The total is " + total;
 }
 
 /* 
--- a/mandatory/2-logic-error.js
+++ b/mandatory/2-logic-error.js
@@ -1,16 +1,18 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
-function getStringLength(word) {
-  return "word".length();
+var sentence = "A wild sentence appeared!";
+var word = "Turtles";
+
+function getStringLength(word, sentence) {
+  return (sentence, word).length;
 }
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  return  a * b * c;;
 }
 
 /* 
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -1,14 +1,18 @@
 // Add comments to explain what this function does. You're meant to use Google!
 function getRandomNumber() {
   return Math.random() * 10;
+  // returns a random number between 1 <1, then multiplies that number by 10
 }
 
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
+  // creates a new array consisting of all elements within the object
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
+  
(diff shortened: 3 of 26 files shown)


mandatory/4-tax.js
```diff
@@ -17,8 +17,12 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
 
+
+
+function addTaxAndFormatCurrency(x) {
+  return (x + 0.20*x).toFixed(2);
```
RitaGlushkova, line 24: Hi Deago, well done for fixing it. If you want to add a pound sign, think of it as a "string" that you add with + in your return formula.

mandatory/3-function-output.js
```diff
@@ -1,14 +1,18 @@
 // Add comments to explain what this function does. You're meant to use Google!
 function getRandomNumber() {
   return Math.random() * 10;
+  // returns a random number between 1 <1, then multiplies that number by 10
```
RitaGlushkova, line 4: Hi Deago, I think there is a typo. Did you mean from 0 to < 1?

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,18 @@
...
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
-function getStringLength(word) {
-  return "word".length();
+var sentence = "A wild sentence appeared!";
+var word = "Turtles";
```
RitaGlushkova, line 8: Your variable is declared correctly. I would recommend using _let_ instead. Because if you work with someone else on a very big piece of code, another person might come up with the same variable name _word_ and without knowing rewrite yours. It will then can then cause a problem.

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,18 @@
...
+
+function getStringLength(word, sentence) {
+  return (sentence, word).length;
 }
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  return  a * b * c;;
```
RitaGlushkova, line 15: I think there is an extra ;

RitaGlushkova (commented):
Hi Deago, as part of our assignment  I decided to review your code. I hope you don't mind. Great work! I like how you assign all functions' calls to variables. In my view, it is a very good practice. An overall note would be about using **let** or **const** instead of **var**.
I think if we developed this practice from the beginning we are less likely to cause bugs in the future. **Var** allows people to completely overwrite it, not just assign a new value. That's why when working in a team, it will be less risky to use **let** or **const**.

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,19 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a, b, c) {
   return a + b + c;
 }
 
-function introduceMe(name, age)
-  return "Hello, my name is " + name "and I am " age + "years old";
+var name = "Sonjide";
```
mahsa2, line 7: This is not needed in the test.

mandatory/3-function-output.js
```diff
@@ -1,14 +1,18 @@
...
 }
 
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
+  // creates a new array consisting of all elements within the object
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
+
```
mahsa2, line 14: No new line needed at the begining of the function.

mandatory/4-tax.js
mahsa2: It can be rewritten like this:
```
return (1.2 * x).toFixed(2);
```

mahsa2 (commented):
Well done on finishing the assignment 💯 👏

PR #245, opened 2021-12-15
London Class 8 - Amarachi Okereke - JavaScript-Core-1-Coursework-Week 1

AmarachiOkereke:

# Your Details

- Your Name: Amarachi Okereke
- Your City: London
- Your Slack Name: Amarachi Okereke

# Homework Details

- Module: JavaScript-Core-1-Coursework
- Week: 1

# Notes

- What did you find easy?
 Writing functions that add, multiply or that takes a string as an input.

- What did you find hard?
 Manipulating functions like calculating percentages, tax etc. It's a bit confusing 

 Some questions relating to functions are still a challenge

- Any other notes?
 I will like to request a follow up teaching on functions please if possible to help me have a better understanding of function.

### DIFF
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -12,8 +12,6 @@ console.log(greet);
 
 function getTotal(a, b) {
   return "The total is " + (a + b);
-
-  
 }
 let result = getTotal(14, 14);
 console.log(result);
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -1,18 +1,22 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a, b, c) {
   return a + b + c;
 }
 
-function introduceMe(name, age)
-  return "Hello, my name is " + name "and I am " age + "years old";
+function introduceMe(name, age) {
+  return "Hello, my name is " + name + " and I am " + age + " years old";
+}
+let greet = introduceMe("Sonjide", 27);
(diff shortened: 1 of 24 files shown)


DelroyGayle (commented):
Dear Amarachi,
I suggest that all your functions should be placed at the top of the program then followed by the code. That is, the definitions of function add(a, b) {} ... function multiply(a, b) ... function format(number) etc should be at the top; then followed by let's, console.log's etc. 
So definitions of functions first, followed by the code that uses them. That generally, is the structured approach to writing programs.

mandatory/1-syntax-errors.js
```diff
@@ -1,18 +1,22 @@
...
 
 function getTotal(a, b) {
-  total = a ++ b;
+  return "The total is " + (a + b);
 
-  return "The total is total";
+  
 }
-
+var result = getTotal(14, 14);
```
DelroyGayle, line 18: 'var' is now considered outdated. Please use 'let' instead

AmarachiOkereke: Yeah, I have corrected it.

AmarachiOkereke:
Thank you Delroy for your review. Can you please indicate where functions are not correctly written in my code, I have checked my code and it seems to be in the correct order. Thank you

> Dear Amarachi, I suggest that all your functions should be placed at the top of the program then followed by the code. That is, the definitions of function add(a, b) {} ... function multiply(a, b) ... function format(number) etc should be at the top; then followed by let's, console.log's etc. So definitions of functions first, followed by the code that uses them. That generally, is the structured approach to writing programs.

PR #266, opened 2021-12-17
london class-8-Nesrin-Sevi-JavaScript-Core-week1

NesrinSevi:

# Your Details

- Your Name:Nesrin
- Your City:London
- Your Slack Name:Nesrin-Trainee

# Homework Details

- Module:JavaScript-core-1
- Week:1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

-----
[View rendered exercises/G-numbers/README.md](https://github.com/NesrinSevi/JavaScript-Core-1-Coursework-Week1/blob/master/exercises/G-numbers/README.md)

### DIFF
--- a/exercises/B-hello-world/exercise.js
+++ b/exercises/B-hello-world/exercise.js
@@ -1,1 +1,1 @@
-console.log("Hello world");
+console.log("Hello world-2");
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,3 +1,3 @@
 // Start by creating a variable `greeting`
-
+var greeting = "Hello World"
 console.log(greeting);
--- a/exercises/D-strings/exercise.js
+++ b/exercises/D-strings/exercise.js
@@ -1,3 +1,7 @@
 // Start by creating a variable `message`
-
+let message = "This is a string";
+let messageType = typeof message;
 console.log(message);
+console.log(messageType);
+
+
--- a/exercises/F-strings-methods/exercise.js
+++ b/exercises/F-strings-methods/exercise.js
@@ -1,3 +1,6 @@
 // Start by creating a variable `message`
 
-console.log(message);
+let name = "Nesrin";
+let nameLength = name.length;
+
+console.log("My name is" + name + " and my name is " + name.lenght + "characters long ");
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,3 +1,5 @@
 const name = " Daniel  ";
+let trimmedName = name.trim()
+let message = "My name is " + trimmedName + " and my name is " + trimmedName.Length + "characters long";
 
 console.log(message);
--- a/mandatory/4-tax.js
+++ b/mandatory/4-tax.js
@@ -5,7 +5,10 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return price * 0.2 + price;
+};
+
 
 /*
   CURRENCY FORMATTING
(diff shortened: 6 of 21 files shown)


exercises/F-strings-methods/exercise.js
```diff
@@ -1,3 +1,6 @@
 // Start by creating a variable `message`
 
-console.log(message);
+let name = "Nesrin";
+let nameLength = name.length;
+
+console.log("My name is" + name + " and my name is " + name.lenght + "characters long ");
```
YDemirer, line 6: I spotted a typo here "name.lenght"

exercises/B-hello-world/exercise.js
```diff
@@ -1 +1 @@
-console.log("Hello world");
+console.log("Hello world-2");
```
gixo, line 1: You may also try to print several things within the same console log, for example: 

console.log(
  "Hello World. I just started learning JavaScript!",
  "Second item",
  99
);

exercises/C-variables/exercise.js
```diff
@@ -1,3 +1,3 @@
 // Start by creating a variable `greeting`
-
+var greeting = "Hello World"
```
gixo, line 2: You should not use var, as this is a legacy keyword. Try to use 'const' if you are not changing the value of the variable, or 'let' if you are changing the content of the variable after the declaration.

exercises/D-strings/exercise.js
```diff
@@ -1,3 +1,7 @@
 // Start by creating a variable `message`
-
+let message = "This is a string";
+let messageType = typeof message;
```
gixo, line 3: you should use `const` instead of `let`, as you are not modifying the value of the variables after initizialisation.

exercises/F-strings-methods/exercise.js
```diff
@@ -1,3 +1,6 @@
 // Start by creating a variable `message`
 
-console.log(message);
+let name = "Nesrin";
+let nameLength = name.length;
```
gixo, line 4: Same consideration here as above, prefer to use `const` rather than `let` whenever possible.

exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,5 @@
 const name = " Daniel  ";
+let trimmedName = name.trim()
+let message = "My name is " + trimmedName + " and my name is " + trimmedName.Length + "characters long";
```
gixo, line 3: When concatenating multiple strings, it is usually preferable to use template literals instead of manual string concatenation.

For example:
```
  return `Hello, my name is ${name} and I am ${age} years old`;
```

mandatory/4-tax.js
```diff
@@ -5,7 +5,10 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return price * 0.2 + price;
```
gixo, line 9: Well done with these functions. A minor improvement to this code would be:

```suggestion
  return price * 1.2;
```

gixo (changes requested):
Well done Nesrin, this is a great submission!

My main remarks are the following: 
 - remember to always declare variables before using them
 - ensure you use `let` and `const` appropriately
 - in general, prefer the use of template literals over manual concatenation

Many of these problems would have been caught by a linter enforcing formatting rules. I suggest you install `eslint` with the Airbnb style guide to see how it works and to study the suggestions that are provided. You can also set up VS Code to automatically fix your code according to the eslint rules.

See this article about installing eslint: https://eslint.org/docs/user-guide/getting-started 

When initializing eslint (`eslint --init`) select the third option (`To check syntax, find problems, and enforce code style`) and then select `Airbnb` for code style presets. You can then install the [jest plugin](https://www.npmjs.com/package/eslint-plugin-jest) for eslint and use the following settings in your `.eslintrc.js` config file.

```
module.exports = {
  env: {
    browser: true,
    es2021: true,
    "jest/globals": true,
  },
  extends: [
    "airbnb-base",
    "plugin:jest/recommended",
  ],
  parserOptions: {
    ecmaVersion: 13,
  },
  rules: {
    "max-len": 0,
    quotes: ["error", "double", { avoidEscape: true }],
    "no-console": "off",
  },
  plugins: ["jest"],
};

```

Once eslint is working, you can configure it to automatically fix your code errors, by following this guide: https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint

I can have another look at the code once you have implemented these changes.

PR #280, opened 2022-05-12
update CI commands to run mandatory and extra tests

lpmi-13:
addresses https://github.com/CodeYourFuture/syllabus/issues/358

I assume the second point was about running the extra tests, though the fix seemed to be different to what was suggested. I'm very happy to amend this PR based on the feedback from somebody more familiar with how the workflow is meant to run.

### DIFF
--- a/.github/workflows/extra-tests.yml
+++ b/.github/workflows/extra-tests.yml
@@ -11,4 +11,4 @@ jobs:
     - name: Install modules
       run: npm install
     - name: Run extra tests
-      run: npm run extra-tests
+      run: npm test -- --selectProjects extra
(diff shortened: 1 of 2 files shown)


SallyMcGrath:
@Dedekind561 There's also this which I think got missed but has just surfaced in my PR lists. 

@moneyinthesky  and @Dedekind561 Please review this while you both have eyes on this week.

Dedekind561 (approved):
LGTM - not seen that `selectProjects` 🎏 before

SallyMcGrath (approved):

SallyMcGrath:
> LGTM - not seen that `selectProjects` 🎏 before

Did you run the build, @Dedekind561 ? I got a broken build when I did. What happened for you?

PR #286, opened 2022-06-29
ZA2 | JavaScript-core1-week1 by Advocate

AdvocateM:
The title for your pull request should be made in this format

Cape-Town-2- Advocate Maroga - JavaScript-Core-1-Week1

# Your Details

— Your Name: Advocate
— Your City: Cape Town
— Your Slack Name: mrmaroga

# Homework Details

- Module: javascript-core-1
- Week: week1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


exercises/J-functions/exercise.js
```diff
@@ -1,5 +1,6 @@
 function halve(number) {
   // complete the function here
+  return number - 6;
 }
 
 var result = halve(12);
```
Xolani-M, line 6: Hey Advocate,

Your code only works in a condition where the argument is equal to 12, when i put halve(24) or any other number except 12 it won't meet the requirement. Consider using a different operator for your code to work in different conditions. 

Thanks

exercises/I-floats/exercise.js
```diff
@@ -1,2 +1,11 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+
+const students = ;
```
AnthonyMogotlane, line 4: You assigned variable students to nothing, when I run this exercise I get a syntax error because of this line.

mandatory/3-function-output.js
```diff
@@ -1,15 +1,18 @@
 // Add comments to explain what this function does. You're meant to use Google!
+// The function round Floats  to the nearest whole number using the `Math.round` function
```
AnthonyMogotlane, line 2: The function below includes Math.random() not Math.round(), the function will return a random number between 0 and less than 10.

AnthonyMogotlane (changes requested):
Keep up the good work Advocate.

exercises/E-strings-concatenation/exercise.js
```diff
@@ -1,3 +1,7 @@
 // Start by creating a variable `message`
+var greetings = "Hello, my name is ";
```
M3ntalist, line 2: Hi Advocate, I see you are still using var to declare variables, as it was mentioned last week, that var is no longer being used much this days and we were advised to use "let" and "const"

AdvocateM: Network destructed me. Thank you, I will change the code

exercises/I-floats/exercise.js
AdvocateM: I forgot to solve it

AdvocateM (commented):
Modified

PR #313, opened 2022-07-31
WM4- Chidimma Ofodum - JavaScript Core 1- Week1

ChidimmaOfodum:

# Your Details

- Your Name: Chidimma Ofodum
- Your City: West Midlands
- Your Slack Name: Chidimma Ofodum

# Homework Details

- Module: JavaScript Core1
- Week: 1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/K-functions-parameters/exercise5.js
+++ b/exercises/K-functions-parameters/exercise5.js
@@ -1,5 +1,9 @@
 // Declare your function here
 
+function createLongGreeting(name , age){
+ return `Hello my name is ${name} and I am ${age} years old`
+}
+
 const greeting = createLongGreeting("Daniel", 30);
 
 console.log(greeting);
--- a/exercises/L-functions-nested/exercise.js
+++ b/exercises/L-functions-nested/exercise.js
@@ -3,3 +3,20 @@ var mentor2 = "Irina";
 var mentor3 = "Mimi";
 var mentor4 = "Rob";
 var mentor5 = "Yohannes";
+
+//  In `exercise.js` you have been provided with the names of some mentors. Write a program that logs a shouty greeting to each one.
+// - Your program should include a function that spells their name in uppercase, and a function that creates a shouty greeting.
+// - Log each greeting to the console.
+
+function mentorGreeting(name) {
+  let getUpperCase = name.toUpperCase();
+  return `Hello ${getUpperCase}`;
+}
+
+console.log(mentorGreeting(mentor1));
+console.log(mentorGreeting(mentor2))
+console.log(mentorGreeting(mentor3));
+console.log(mentorGreeting(mentor4));
+console.log(mentorGreeting(mentor5));
+
+
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -1,16 +1,16 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a,b,c) {
   return a + b + c;
 }
 
-function introduceMe(name, age)
-  return "Hello, my name is " + name "and I am " age + "years old";
+function introduceMe(name, age){
+  return `Hello, my name is ${name} and I am ${age} years old`;
+}
 
 function getTotal(a, b) {
-  total = a ++ b;
-
-  return "The total is total";
+  let total = a + b;
+  return `The total is ${total}`;
 }
 
 /* 
--- a/mandatory/2-logic-error.js
+++ b/mandatory/2-logic-error.js
@@ -1,16 +1,16 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 function getStringLength(word) {
-  return "word".length();
+ return word.length;
 }
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  
+  return a * b * c;
 }
 
 /* 
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -1,9 +1,11 @@
-// Add comments to explain what this function does. You're meant to use Google!
+
+//This function returns a random number between 0 and 10
 function getRandomNumber() {
   return Math.random() * 10;
 }
 
-// Add comments to explain what this function does. You're meant to use Google!
+
+//This function joins word1 and word2 together
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
@@ -11,6 +13,8 @@ function combine2Words(word1, word2) {
 function concatenate(firstWord, secondWord, thirdWord) {
   // Write the body of this function to concatenate three words together.
   // Look at the test case below to understand what this function is expected to return.
+
+  return firstWord.concat(" ", secondWord, " ", thirdWord)
 }
 
 /* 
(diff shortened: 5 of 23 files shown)


Alex-Phillip:
exercises/K-functions-parameters/exercise5.js

exercises/K-functions-parameters/exercise5.js
```diff
@@ -1,5 +1,9 @@
 // Declare your function here
 
+function createLongGreeting(name , age){
+ return `Hello my name is ${name} and I am ${age} years old`
```
Alex-Phillip, line 4: For example line 4 here, so simple. Mine looked a lot more complicated than that to get the same result. I will try to remember how you did this and use the same method in future. Thanks!

exercises/L-functions-nested/exercise.js
```diff
@@ -3,3 +3,20 @@ var mentor2 = "Irina";
...
 var mentor4 = "Rob";
 var mentor5 = "Yohannes";
+
+//  In `exercise.js` you have been provided with the names of some mentors. Write a program that logs a shouty greeting to each one.
+// - Your program should include a function that spells their name in uppercase, and a function that creates a shouty greeting.
+// - Log each greeting to the console.
+
+function mentorGreeting(name) {
+  let getUpperCase = name.toUpperCase();
+  return `Hello ${getUpperCase}`;
```
AzinYad, line 13: very well usage of  nested function

jaspreet338 (commented):
hello chiddima, i like  yours exercise.js from E-strings-concatenation because i find it easier way than 

[](https://github.com/CodeYourFuture/JavaScript-Core-1-Coursework-Week1/pull/313/files#diff-f8216b9daa3e5e38f93d56e2b7685e1402accee3b2e46731abd1f452b350d75f)

exercises/K-functions-parameters/exercise5.js
ChidimmaOfodum: You are welcome Alex!

exercises/L-functions-nested/exercise.js
ChidimmaOfodum: Thank you Azin

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,16 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a,b,c) {
   return a + b + c;
 }
```
alicagatay, line 5: That is correct, when defining parameters in a function, you should always separate the parameters with a `,`

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,16 @@
...
 
-function introduceMe(name, age)
-  return "Hello, my name is " + name "and I am " age + "years old";
+function introduceMe(name, age){
+  return `Hello, my name is ${name} and I am ${age} years old`;
+}
```
alicagatay, line 9: That is right, after writing your function definition, you should always surround the body of your function with curly brackets, otherwise, JavaScript won't understand that the body you wrote belongs to that function.

I also liked how you used String interpolation in this part.

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,16 @@
...
 
 function getTotal(a, b) {
-  total = a ++ b;
-
-  return "The total is total";
+  let total = a + b;
+  return `The total is ${total}`;
 }
```
alicagatay, line 14: That's right. Ass you see, you cannot use `++` to add to String values together, `++` can only be used while increasing the value of a variable by 1. Other than that, as you did on this question, you always need to use either the let or the const keyword while defining a variable, otherwise JavaScript won't understand that it is a variable, and won't save it on its memory.

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,16 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
```
alicagatay, line 5: That's right, there is no such function in JavaScript called `wordtrim()`. You should use `trim()` instead.

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,16 @@
...
 
 function getStringLength(word) {
-  return "word".length();
+ return word.length;
 }
```
alicagatay, line 9: That is correct, you cannot surround a variable with quotation mark, since JavaScript will treat it as a String.

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,16 @@
...
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  
+  return a * b * c;
 }
```
alicagatay, line 14: That is right.

mandatory/3-function-output.js
```diff
@@ -1,16 +1,20 @@
-// Add comments to explain what this function does. You're meant to use Google!
+
+//This function returns a random number between 0 and 10
 function getRandomNumber() {
   return Math.random() * 10;
 }
```
alicagatay, line 5: "The Math.random() function returns a floating-point, pseudo-random number in the range 0 to less than 1 (inclusive of 0, but not 1) with approximately uniform distribution over that range — which you can then scale to your desired range."

The above description is taken from MDN Docs. According to this, the `Math.random()` function randomly generates a number between 0 and 1, then multiplies it with a number of your choice.

mandatory/3-function-output.js
```diff
@@ -1,16 +1,20 @@
...
 
-// Add comments to explain what this function does. You're meant to use Google!
+
+//This function joins word1 and word2 together
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
```
alicagatay, line 11: That's true, the `concat()` function will add `word2` into `word1`.

mandatory/3-function-output.js
```diff
@@ -1,16 +1,20 @@
...
 
 function concatenate(firstWord, secondWord, thirdWord) {
   // Write the body of this function to concatenate three words together.
   // Look at the test case below to understand what this function is expected to return.
+
+  return firstWord.concat(" ", secondWord, " ", thirdWord)
 }
```
alicagatay, line 18: That's right.

mandatory/4-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(amount) {
+  return amount * 1.2;
+}
```
alicagatay, line 10: That's correct.

mandatory/4-tax.js
```diff
@@ -17,7 +19,12 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price) {
+  let formatPrice = (calculateSalesTax(price)).toFixed(2);
+  return `£${formatPrice}`;
+}
```
alicagatay, line 25: That's also correct.

alicagatay (changes requested):
Good work Chidimma. I really liked the way you have used String interpolation in your answers. Keep up the good work.

PR #351, opened 2022-08-05
WM_4 - ALESSANDRA_SOUSA - JAVASCRIPT - WEEK 1

AlessandraSousa:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/L-functions-nested/exercise.js
+++ b/exercises/L-functions-nested/exercise.js
@@ -1,5 +1,19 @@
-var mentor1 = "Daniel";
-var mentor2 = "Irina";
-var mentor3 = "Mimi";
-var mentor4 = "Rob";
-var mentor5 = "Yohannes";
+
+function greetings(mentor) {
+let mentor1 = "Daniel";
+let mentor2 = "Irina";
+let mentor3 = "Mimi";
+let mentor4 = "Rob";
+let mentor5 = "Yohannes";
+let result = `Hello ${mentor}`
+console.log(result.toUpperCase())
+}
+
+greetings("Daniel");
+greetings("Irina");
+greetings("Mimi");
+greetings("Rob");
+greetings("Yohannes");
+   
+    
+    
--- a/mandatory/2-logic-error.js
+++ b/mandatory/2-logic-error.js
@@ -1,16 +1,16 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return trimWord();
 }
 
 function getStringLength(word) {
-  return "word".length();
+  return word.length();
 }
 
 function multiply(a, b, c) {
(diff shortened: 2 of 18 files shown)


exercises/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,19 @@
...
-var mentor3 = "Mimi";
-var mentor4 = "Rob";
-var mentor5 = "Yohannes";
+
+function greetings(mentor) {
+let mentor1 = "Daniel";
+let mentor2 = "Irina";
+let mentor3 = "Mimi";
+let mentor4 = "Rob";
+let mentor5 = "Yohannes";
```
jcholyhead, line 7: These variables don't do anything here, because you are passing the strings directly into the mentor parameter.

mandatory/3-function-output.js
```diff
@@ -1,16 +1,19 @@
...
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
+  //This function joins two words together. 
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
   // Write the body of this function to concatenate three words together.
   // Look at the test case below to understand what this function is expected to return.
+  return word.concat(firstWord, secondWord, thirdWord);
```
jcholyhead, line 16: I don't think this will work - what is `word`?

The concat function is called on a string. You can solve this exercise with this function, but it will look (something) like this:

`return firstWord.concat(" ", secondWord, " ", thirdWord);`

Better to use interpolation: 
```
return `${firstWord} ${secondWord} ${thirdWord}`;

**Review by jcholyhead — COMMENTED** [review]
A good rule of thumb for JS "variables". Always use the `const` keyword, unless the variable must change later in the script. This small change will make your code look a lot more 'professional' and will make it stand out during a job application process.

File: exercises/L-functions-nested/exercise.js (hunk 1 above)
Comment (AlessandraSousa) [reply exercises/L-functions-nested/exercise.js:7]: Corrected, thank you.

**Comment by AlessandraSousa** [comment]
After trying to correct the errors.

File: mandatory/2-logic-error.js [hunk 3]
```diff
@@ -1,16 +1,16 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 function getStringLength(word) {
-  return "word".length();
+  return word.length();
```
Pat-On, line 8: You do not need to use () when you call length(). It should be `word.length`, without a function call. 
Please check this: https://www.w3schools.com/jsref/jsref_length_string.asp

PR #356, opened 2022-08-05
JavaScript Challenges and Exercises

ghost:
West Midlands_Class 4_Driss Hmaidi_JavaScript Coursework_Week 1

### DIFF
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -11,6 +11,7 @@ function combine2Words(word1, word2) {
 
 // The function will return the following "firstWord, secondWord, thirdWord" 
 function concatenate(firstWord, secondWord, thirdWord) {
+  return firstWord + " " + secondWord + " " + thirdWord;
 }
 
 
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -1,17 +1,18 @@
-// Add comments to explain what this function does. You're meant to use Google!
+//The getRandomNumber function will return a random integer from 0 to 9.
+
 function getRandomNumber() {
   return Math.random() * 10;
 }
 
-// Add comments to explain what this function does. You're meant to use Google!
+// The combine2Words function will combine the two words and create a new string.
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
 
+// The function will return the following "firstWord, secondWord, thirdWord" 
 function concatenate(firstWord, secondWord, thirdWord) {
-  // Write the body of this function to concatenate three words together.
-  // Look at the test case below to understand what this function is expected to return.
 }
+
 
 /* 
 ===================================================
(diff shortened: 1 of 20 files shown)


mandatory/3-function-output.js
```diff
@@ -1,18 +1,20 @@
-// Add comments to explain what this function does. You're meant to use Google!
+//The getRandomNumber function will return a random integer from 0 to 9.
+
 function getRandomNumber() {
   return Math.random() * 10;
 }
```
alicagatay, line 5: That is correct. The `Math.random()` function returns a number between 0 and 1(inclusive of 0, but not including 1), and then it will just multiply that value by 10 as we wrote `* 10` after the method.

mandatory/3-function-output.js
```diff
@@ -1,18 +1,20 @@
...
 
-// Add comments to explain what this function does. You're meant to use Google!
+// The combine2Words function will combine the two words and create a new string.
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
```
alicagatay, line 10: That's right, the `combine2Words()` function will combine two strings with the help of the `concat()` method.

mandatory/3-function-output.js
```diff
@@ -1,18 +1,20 @@
...
 
+// The function will return the following "firstWord, secondWord, thirdWord" 
 function concatenate(firstWord, secondWord, thirdWord) {
-  // Write the body of this function to concatenate three words together.
-  // Look at the test case below to understand what this function is expected to return.
+  return firstWord + " " + secondWord + " " + thirdWord;
 }
```
alicagatay, line 15: Again, nice way of concatenating multiple words.

mandatory/4-tax.js
```diff
@@ -5,7 +5,13 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+return price * 1.2;
+}
```
alicagatay, line 10: That is correct.

You could also do something lie `const amountOfTax = originalPrice * (taxRate / 100);` to calculate the tax and then add that tax to the original price, but your solution is also correct.

mandatory/4-tax.js
```diff
@@ -17,7 +23,12 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price) {
+  let finalPrice = (price *1.2).toFixed(2)
```
alicagatay, line 27: On the part `(price * 1.2)`, you can also use the `calculateSalesTax()` function you just wrote to make things a bit simpler.

alicagatay (changes requested):
Nice work Driss, I can clearly see that you understood this week of JavaScript - 1 very well.

Keep up the good work.

PR #357, opened 2022-08-06
WM4_Olivier_Sedzikowski_JS1_week_1

OlivierSedzikowski:

# Your Details

- Your Name:Olivier Sedzikowski
- Your City:Birmingham
- Your Slack Name:olivier sedzikowski

# Homework Details

- Module: JS1
- Week: week 1

# Notes

- What did you find easy?
- Quiet hard.

- What did you find hard?
- Making function.

- How to test my codes.

- Any other notes?
- If we could have access to the solutions to have an idea of what is right because we code but we don't know what is right or wrong.

### DIFF
--- a/mandatory/2-logic-error.js
+++ b/mandatory/2-logic-error.js
@@ -1,16 +1,15 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return trimWord();
 }
 
 function getStringLength(word) {
-  return "word".length();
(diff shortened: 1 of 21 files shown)


mandatory/2-logic-error.js
```diff
@@ -1,16 +1,15 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return trimWord();
```
irinamihai150, line 4: You could try to use  - return word.trim();

PR #361, opened 2022-08-09
wm4class2_luleka mgaga_javascript_week1

LMgaga:

# Your Details

- Your Name: Luleka
- Your City: Birmingham
- Your Slack Name: Luleka Mgaga

# Homework Details

- Module:
- Week: 
wm4class2_luleka mgaga_javascript_week1
# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/2-logic-error.js
+++ b/mandatory/2-logic-error.js
@@ -1,16 +1,16 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 function getStringLength(word) {
-  return "word".length();
+  return word.length;
 }
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  return a * b * c;
+  
 }
 
 /* 
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -9,6 +9,7 @@ function combine2Words(word1, word2) {
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
+  return firstWord + ' ' + secondWord + ' ' + thirdWord
   // Write the body of this function to concatenate three words together.
   // Look at the test case below to understand what this function is expected to return.
 }
--- a/mandatory/4-tax.js
+++ b/mandatory/4-tax.js
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return price + (price * (20 / 100 ))
+}
 
 /*
   CURRENCY FORMATTING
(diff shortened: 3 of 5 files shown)


mandatory/2-logic-error.js
```diff
@@ -1,16 +1,16 @@
...
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 function getStringLength(word) {
-  return "word".length();
+  return word.length;
 }
```
alicagatay, line 9: Yeap that's right, you cannot wrap a variable with `""` as JavaScript would think it is a String value if you did.

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,16 @@
...
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  return a * b * c;
+  
 }
```
alicagatay, line 14: That's true, the `return` keyword and your operation must be in the same line. Or, another thing you can do is you can define a variable and and make it equal to `a * b * c`. After doing that, you can directly return the variable.

mandatory/3-function-output.js
```diff
@@ -9,6 +9,7 @@ function combine2Words(word1, word2) {
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
+  return firstWord + ' ' + secondWord + ' ' + thirdWord
```
alicagatay, line 12: That's right, you can concatenate multiple String variables simply by how you did.

mandatory/4-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  return price + (price * (20 / 100 ))
+}
```
alicagatay, line 10: That's correct, in order to calculate the tax, you can multiply the value by 20 and then divide it by 100. And then to find the final price, you can add it to the original price.

mandatory/4-tax.js
```diff
@@ -17,7 +19,11 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price) {
+  tax = calculateSalesTax(price);//calculating tax
+  //final_price = price + tax; //calculating final price
+ return `£${tax.toFixed(2)}`; 
+}
```
alicagatay, line 26: That is right, you can easily format your value by using String interpolation, and you can use the function `toFixed()` to convert a number into specified number of digits.

alicagatay (changes requested):
Nice work Luleka, I can really see that you know the basics very well. Keep up the good work!

PR #371, opened 2022-08-22
Manchester NW5-LeilaFaez-Js-Core1-Week1

leilafaez:

# Your Details

- Your Name:Leila Faez
- Your City:Manchester
- Your Slack Name:leilafaez

# Homework Details

- Module:Javascript
- Week:1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -1,16 +1,18 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a, b, c) {
   return a + b + c;
 }
 
-function introduceMe(name, age)
-  return "Hello, my name is " + name "and I am " age + "years old";
+function introduceMe(name, age){
+  return "Hello, my name is " + name + " and I am " + age + " years old";
+}
+
 
 function getTotal(a, b) {
-  total = a ++ b;
+  total = a + b;
 
-  return "The total is total";
+  return "The total is " +total;
 }
 
 /* 
--- a/mandatory/2-logic-error.js
+++ b/mandatory/2-logic-error.js
@@ -1,16 +1,22 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  
+  return word.trim();
 }
+word = "  CodeYourFuture  ";
+word = " CodeYourFuture teaches coding     ";
 
 function getStringLength(word) {
-  return "word".length();
+ 
+  return word.length;
 }
+ word = getStringLength("A wild sentence appeared!");
+ word = getStringLength("Turtle"); ;
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+ const result= a * b * c;
+  return result;
 }
 
 /* 
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -1,17 +1,24 @@
 // Add comments to explain what this function does. You're meant to use Google!
+
+/* This function creat an integer random number(by multiplying decimal random number by ten)*/
 function getRandomNumber() {
   return Math.random() * 10;
 }
 
 // Add comments to explain what this function does. You're meant to use Google!
+
+/*This function concat two words which will be pass to the function as an argument
+for example if word1="ali" and word="reza" the result of this function will be "alireza"*/
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
-  // Write the body of this function to concatenate three words together.
-  // Look at the test case below to understand what this function is expected to return.
+  return firstWord + " " + secondWord + " " + thirdWord;
 }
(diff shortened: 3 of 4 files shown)


mandatory/3-function-output.js
```diff
@@ -1,17 +1,24 @@
...
+/*This function concat two words which will be pass to the function as an argument
+for example if word1="ali" and word="reza" the result of this function will be "alireza"*/
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
-  // Write the body of this function to concatenate three words together.
-  // Look at the test case below to understand what this function is expected to return.
+  return firstWord + " " + secondWord + " " + thirdWord;
```
levilu800b, line 17: Hello Leila,  you have done an amazing job and your codes are simple and easy to understand and review. However, I think for the return you could have used .concat to concatenate all the words together like on the above combine2Words function. If there is a reason for not using that do you mind explaining? Thank you

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,18 @@
...
+}
+
 
 function getTotal(a, b) {
-  total = a ++ b;
+  total = a + b;
 
-  return "The total is total";
+  return "The total is " +total;
 }
```
Meronggk, line 16: You have solved the exercises and mandatory problems in a very neat, simple and understandable code and it's great. Well done.

mandatory/4-tax.js
```diff
@@ -17,7 +22,10 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price) {
+ 
+  return  `£${calculateSalesTax(price).toFixed(2)}`;
```
levilu800b, line 27: Love the way you used what was taught in class, nice work.

levilu800b (commented):
I really enjoyed reviewing your code. The simplicity and the indentation make it easier for people or anyone else to read and understand.

mandatory/1-syntax-errors.js
leilafaez: > 

Many thanks :)

mandatory/3-function-output.js
leilafaez: Thank you for your feedback. Yes you are completely right. I just wanted to try every piece of code that we have learnt during course :)

extra/3-magic-8-ball.js
```diff
@@ -58,10 +88,25 @@ function shakeBall() {
...
+   let ballPredict = fixAnswer.indexOf(answer);
+   if (ballPredict <= 4) {
+     answer = "very positive";
+   } else if (ballPredict <= 9) {
+     answer = "positive";
+   } else if (ballPredict <= 14) {
+     answer = "negative";
+   } else if (ballPredict <= 19) {
+     answer = "very negative";
+   } else answer = "Something broken";
```
Gevie, line 103: Hi Leila, you've pretty much hit the nail on the head here 👍 

I would just be careful at the end with your `} else answer = "Something broken";` as this syntax isn't quite correct. I also notice you define answer then return it at the end, another way of doing this is just to return in each conditional block. For example:

```javascript
  const ballPredict = fixAnswer.indexOf(answer);
  
  if (ballPredict <= 4) {
    return "very positive";
  } else if (ballPredict <= 9) {
    return "positive";
  } else if (ballPredict <= 14) {
    return "negative";
  } else if (ballPredict <= 19) {
    return "very negative";
  }
  
  return "Something broken";
}
```

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,22 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  
+  return word.trim();
 }
+word = "  CodeYourFuture  ";
```
Gevie, line 7: You have successfully completed the exercise by fixing the errors 💯 

We are however defining a `word` variable 4 times in this file, reassigning its value and never actually using the variable anywhere. We can probably delete these lines :)

mandatory/3-function-output.js
```diff
@@ -1,17 +1,24 @@
...
 }
+firstWord="code";
```
Gevie, line 19: You've completed this mandatory exercise spot on Leila :)

The only feedback would be that we don't need the `firstWord`, `secondWord` and `thirdWord` variables on line 19 to 21 since they're never used 👍

mandatory/4-tax.js
```diff
@@ -4,8 +4,13 @@
   A business requires a program that calculates how much the price of a product is including sales tax
   Sales tax is 20% of the price of the product.
 */
+ 
+ 
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  const saleTax = (price * 20) / 100;
```
Gevie, line 11: Great job, there's nothing I can suggest for improvements here :)

PR #373, opened 2022-08-22
NW5_Levi_Udeh_JavaScript_1_Week_1

levilu800b:

# Your Details

- Your Name: Levi Udeh
- Your City: Leeds
- Your Slack Name: Levi

# Homework Details

- Module: JS
- Week: 1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/B-hello-world/exercise.js
+++ b/exercises/B-hello-world/exercise.js
@@ -1,1 +1,2 @@
-console.log("Hello world");
+console.log('Hello world');
+console.log('Hello world');


Ekremteke (commented):
Good job @levilu800b . Your work has passed all of the tests and you have a green thick which I don't have. One thing that I learned today is  we should use "let" and "const" instead of 
"var".

levilu800b:
> Good job @levilu800b . Your work has passed all of the tests and you have a green thick which I don't have. One thing that I learned today is we should use "let" and "const" instead of "var".

Thanks for the feedback. I will work on that and avoid using Var next time. Thank you

PR #385, opened 2022-08-25
Change and solve mandatory tasks.

Mohammadrezanafar:

# Your Details

- Your Name: Mohammadreza Nafar
- Your City: Northwest
- Your Slack Name: Mohammadreza

# Homework Details

- Module: Javascript
- Week: Week1

# Notes

- What did you find easy? exercise was easy

- What did you find hard? To start, it was a little hard to understand the task. 

- What do you still not understand? I should have more study about function part to understand it better.

- Any other notes? no

### DIFF
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,3 +1,7 @@
 // Start by creating a variable `greeting`
 
+
+var greeting="hello world";
+console.log(greeting);
+console.log(greeting);
 console.log(greeting);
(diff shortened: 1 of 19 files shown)


Barryeasy90:
Hi @Mohammadrezanafar 
 I can see that your code is easy to read and using very few variable which is very nice. However I will suggest to avoid using var to declare a variable as this is a old version instead using let and also avoid using very short variable name like a, b or x instead using a very generic name, what the variable or parameter does.

#happycoding

PR #388, opened 2022-08-26
NW5-Manchester-Sandra Duarte-JavaScript-Week1

Sandra-Duarte:
- Your Name: Sandra Duarte
- Your City: NW5-Manchester
- Your Slack Name: Sandra Duarte

# Homework Details

- Module: JavaScript
- Week: 1

# Notes

- What did you find easy?

- What did you find hard?

- Functions-parameters

- Any other notes?

### DIFF
--- a/exercises/B-hello-world/exercise.js
+++ b/exercises/B-hello-world/exercise.js
@@ -1,1 +1,3 @@
 console.log("Hello world");
+console.log("Hello world I'm sandra")
+console.log("I'm learning JavScript")
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,4 +1,5 @@
 // Start by creating a variable `greeting`
-
+const greeting = "How are you doing."
 console.log(greeting);
-console.log("Hello World")
+console.log(greeting)
+
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,3 +1,4 @@
 // Start by creating a variable `greeting`
 
 console.log(greeting);
+console.log("Hello World")
--- a/exercises/D-strings/exercise.js
+++ b/exercises/D-strings/exercise.js
@@ -1,3 +1,4 @@
 // Start by creating a variable `message`
-
+const message = "This is a string.";
 console.log(message);
+console.log(typeof message);
--- a/exercises/E-strings-concatenation/exercise.js
+++ b/exercises/E-strings-concatenation/exercise.js
@@ -1,3 +1,7 @@
 // Start by creating a variable `message`
+const startGreeting = "Hi my name is Sandra.";
+const name = "Sandra Duarte";
+
 
 console.log(message);
+console.log(startGreeting + " " + name);


exercises/K-functions-parameters/exercise.js
```diff
@@ -1,6 +1,10 @@
 // Complete the function so that it takes input parameters
 function multiply() {
   // Calculate the result of the function and return it
+  function multiply(a, b) {
```
Ara225, line 4: This won't work properly, the outer function is the one that's getting called as the inner function is scoped to the inside of the curly brackets it's in

In the sample below result will equal 12 cos it's being called within the same curly brackets the inner function is defined in and the inner function is overriding the outer function
```
function multiply() {
  // Calculate the result of the function and return it
  function multiply(a, b) {
  return a*b;
  }
  let result = multiply(3, 4);
}
```
But it will be undefined in this case as it only sees the outer function which doesn't return anything
```
function multiply() {
  // Calculate the result of the function and return it
  function multiply(a, b) {
  return a*b;
  }
  
}
let result = multiply(3, 4);
```

exercises/K-functions-parameters/exercise2.js
```diff
@@ -1,4 +1,7 @@
 // Declare your function first
+function divide(num1, num2){
+return num1/num2;
```
Ara225, line 3: bit of improperly indented code here, can be a pain futher down the line so it's best to get into good habits now. The Prettier exstention should be doing the work for you see https://syllabus.codeyourfuture.io/guides/code-style-guide#indent-your-code

exercises/K-functions-parameters/exercise4.js
```diff
@@ -1,5 +1,10 @@
 // Declare your function first
+function sum2numbers (num1, num2){
+return num1 + num2
```
Ara225, line 3: Another improperly indented code piece

extra/1-currency-conversion.js
```diff
@@ -16,6 +19,11 @@ function convertToUSD() {}
 */
 
 function convertToBRL() {}
+function convertToUSD(value){
```
Ara225, line 22: You've got three functions with the same name here, this will cause a syntax error

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,23 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
 function addNumbers(a b c) {
-  return a + b + c;
+  return a + b + c;}
+  function addNumbers(a, b, c) {
```
Ara225, line 5: Afraid this won't work as you'd think see my notes about the multiply function in exercises/K-functions-parameters/exercise.js If there's anything you don't understand in it let me know

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,23 @@
...
+    return  a  +  b  +  c ;
 }
 
 function introduceMe(name, age)
   return "Hello, my name is " + name "and I am " age + "years old";
+  function introduceMe(name, age) {
```
Ara225, line 11: As above

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,23 @@
...
+    return "Hello, my name is " + name + " and I am " + age + " years old";
+  }
 
 function getTotal(a, b) {
   total = a ++ b;
+  total = a + b;
 
   return "The total is total";
+  return "The total is " + total;
```
Ara225, line 20: This line of code will never be ran. If you look at it in VsCode, it's slightly lighter than the code around it so you can easily tell that it's redundant

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,23 @@
...
 
 function introduceMe(name, age)
   return "Hello, my name is " + name "and I am " age + "years old";
+  function introduceMe(name, age) {
+    return "Hello, my name is " + name + " and I am " + age + " years old";
+  }
 
 function getTotal(a, b) {
   total = a ++ b;
+  total = a + b;
```
Ara225, line 17: Right answer, but the line above should be deleted - that causes a syntax error

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,16 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 function getStringLength(word) {
   return "word".length();
+  return word.trim();
```
Ara225, line 9: This code will never be reached

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,16 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 function getStringLength(word) {
   return "word".length();
```
Ara225, line 8: Throws error

mandatory/4-tax.js
```diff
@@ -6,6 +6,9 @@
 */
 
 function calculateSalesTax() {}
+function calculateSalesTax(price) {
```
Ara225, line 9: duplicate functions again

mandatory/4-tax.js
```diff
@@ -18,7 +21,9 @@ function calculateSalesTax() {}
 */
 
 function addTaxAndFormatCurrency() {}
-
+function addTaxAndFormatCurrency(price) {
```
Ara225, line 24: duplicate functions

mandatory/4-tax.js
```diff
@@ -32,24 +37,30 @@ To run the tests for just this one file, type `npm test -- --testPathPattern 4-t
 
 test("calculateSalesTax for £15", () => {
   expect(calculateSalesTax(15)).toEqual(18);
+  expect(calculateSalesTax(15)).toEqual(18);
```
Ara225, line 40: Not sure why you've put extra expect calls in this file? They're unneeded

Ara225 (commented):
Good work! I see you're struggling a little with functions but I think a lot of it may be down to fairly simple misunderstandings/mistakes, I've called them out when I've noticed and tried to explain, hopefully that causes that to click in your head. I see mandatory/3-function-output.js got overlooked by the way

Ara225:
Noticed the descriptive commit names which are a nice touch Best to do one task per commit if possible :)

PR #393, opened 2022-08-26
north-west-5/kam-nik

KamNik0044:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


SteveLeicester:
Thanks for submitting week one JS
Just a comment on the exercises - it is generally better to declare variables with `let` rather than` var`. Though you will continue to see `var `in older code modern practice is to use `let`.  Certainly try to avoid mixing both together also.

The mandatory tests look good and pass so well done there.

Steve

PR #411, opened 2022-11-19
London 9 - Baz Murphy - JavaScript Core 1 - Week 1

bazmurphy:
# Your Details

- Your Name: Baz Murphy 
- Your City: London
- Your Slack Name: Baz

# Homework Details

- Module: JavaScript Core 1
- Week: 1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/2-piping.js
+++ b/extra/2-piping.js
@@ -16,26 +16,56 @@
   the final result to the variable goodCode
 */
 
-function add() {
+// function add() {
 
+// }
+
+function add(numberOne, numberTwo) {
+  return numberOne + numberTwo;
 }
 
-function multiply() {
+// const add = (numberOne, numberTwo) => numberOne + numberTwo;
 
+// function multiply() {
+
+// }
+
+function multiply(numberOne, numberTwo) {
+  return numberOne * numberTwo;
 }
 
-function format() {
+// const multiply = (numberOne, numberTwo) => numberOne * numberTwo;
 
+// function format() {
+
+// }
+
+function format(number) {
+  return `£${number}`
 }
+
+// const format = (number) => `£${number}`
 
 const startingValue = 2;
 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+let badCode = format(multiply(add(startingValue, 10), 2));
+// this is calling 3 functions 2 of which are nested, its bad practice because its unreadable(?) and not easy to maintain(?)
+
+console.log(badCode);
+// £24
 
 /* BETTER PRACTICE */
 
-let goodCode =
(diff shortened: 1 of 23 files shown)


extra/2-piping.js
```diff
@@ -16,26 +16,56 @@
...
 }
 
+// const format = (number) => `£${number}`
+
 const startingValue = 2;
 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+let badCode = format(multiply(add(startingValue, 10), 2));
+// this is calling 3 functions 2 of which are nested, its bad practice because its unreadable(?) and not easy to maintain(?)
```
PeteLindsell, line 53: Yes the bad practice is the readability.

extra/2-piping.js
```diff
@@ -16,26 +16,56 @@
...
+
+console.log(badCode);
+// £24
 
 /* BETTER PRACTICE */
 
-let goodCode =
+const pipe = (functionOne, functionTwo, functionThree) => (functionOneParameterOne, functionOneParameterTwo, functionTwoParameterTwo) => functionThree(functionTwo(functionOne(functionOneParameterOne, functionOneParameterTwo), functionTwoParameterTwo));
+let goodCode = pipe(add, multiply, format)(startingValue, 10, 2);
+// i don't neccessarily agree that pipe/piping is good code...
```
PeteLindsell, line 62: I agree this is much less readable.

A pipe function can be good when you have multiple functions that take one argument that is the output of the last but as you can see from above its not the best when the functions take multiple arguments.

I think what we were after here is something more simple like this:
```
const sum = add(startingValue, 10);
const product = multiply(sum, 2);
const goodCode = format(product);
```

The tricky thing is thinking of good names

extra/3-magic-8-ball.js
```diff
@@ -1,3 +1,8 @@
+// import jest extended matchers
+const jestExtended = require("jest-extended");
+// extend the existing jest class with the jest-extended matchers
+expect.extend(jestExtended);
+
```
PeteLindsell, line 5: Well done for finding a fix for this

PeteLindsell (commented):
Really good work! I can see you understand the concepts and have explained them well in the comments.

extra/2-piping.js
bazmurphy: When I was researching piping it was all about how you make an overall piping function that handles everything, 
and how you are then able to use the two sets of "parameter" parentheses back to back... (which I had not encountered up until that point)

That's why I did it the way I did. After learning how to do it like that^

I think that is far more difficult to wrap my head around than the example you gave of what they were looking for

So is my understanding of what piping is incorrect?😕❓ 
Is the example you gave also "piping" and whats the difference between the two? 🤔💭

extra/3-magic-8-ball.js
bazmurphy: They need to fix this on the repository itself! 👿

PR #412, opened 2022-11-20
London 9 - susan - JavaScript Core 1 - Week 1

hkuhku609:


OlushoreOdude:
Hi Susan,
Great use of chaining methods, makes the code clear, concise and easy to read. Thanks Olus

mandatory/3-function-output.js
```diff
@@ -1,16 +1,19 @@
 // Add comments to explain what this function does. You're meant to use Google!
 function getRandomNumber() {
-  return Math.random() * 10;
+  return Math.random() * 10
+  //The Math.random() function returns a floating-point, pseudo-random number that's greater than or equal to 0 and less than 1
```
samhoooo, line 4: Correct. And that number is multiply by 10, so the function returns a number greater than 0 and smaller than 10.

samhoooo (commented):
Nicely done! I have only got a minor comment through, look good

PR #426, opened 2022-11-22
LONDON9-LOVELACE-BOSHRA-MAHMOUDI-JS1-WEEK1

BoshraM:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,3 +1,5 @@
 // Start by creating a variable `greeting`
-
+var greeting = "I am BOshra";
+console.log(greeting);
+console.log(greeting);
 console.log(greeting);
--- a/exercises/E-strings-concatenation/exercise.js
+++ b/exercises/E-strings-concatenation/exercise.js
@@ -1,3 +1,5 @@
 // Start by creating a variable `message`
-
+let greeting = "My name is";
+let myName = "Boshra";
+let message = greeting + " " + myName;
 console.log(message);
--- a/exercises/I-floats/exercise.js
+++ b/exercises/I-floats/exercise.js
@@ -1,2 +1,8 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+let totalNumber = numberOfMentors + numberOfStudents;
+let percentageOfStudents = (numberOfStudents / totalNumber) * 100;
+let roughPercentageOfStudents = Math.round(percentageOfStudents);
+console.log(roughPercentageOfStudents);
+let percentageOfMentors = (numberOfMentors / totalNumber) * 100;
+console.log(Math.round(percentageOfMentors));
--- a/exercises/K-functions-parameters/exercise3.js
+++ b/exercises/K-functions-parameters/exercise3.js
@@ -1,5 +1,11 @@
 // Write your function here
+function createGreeting(string) {
+  return "my name is " + string;
+}
 
 var greeting = createGreeting("Daniel");
 
+console.log(greeting);
+
+var greeting = createGreeting("Boshra");
 console.log(greeting);
--- a/exercises/L-functions-nested/exercise.js
+++ b/exercises/L-functions-nested/exercise.js
@@ -3,3 +3,17 @@ var mentor2 = "Irina";
 var mentor3 = "Mimi";
 var mentor4 = "Rob";
 var mentor5 = "Yohannes";
+
+function upperCaseName(mentor) {
+  return mentor.toUpperCase();
+}
+
+function greeting(mentor) {
+  return `HELLO ${upperCaseName(mentor)}`;
+}
+
+console.log(greeting(mentor1));
+console.log(greeting(mentor2));
+console.log(greeting(mentor3));
+console.log(greeting(mentor4));
+console.log(greeting(mentor5));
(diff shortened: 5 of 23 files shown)


exercises/C-variables/exercise.js
```diff
@@ -1,3 +1,5 @@
 // Start by creating a variable `greeting`
-
+var greeting = "I am BOshra";
```
Dedekind561, line 2: Hi there @BoshraM, 
We've just updated our coursework with a change to the variable declarations.
From now on when you're writing variable declarations be sure to use `let` / `const` over `var`.
```suggestion
const greeting = "I am BOshra";
```

exercises/E-strings-concatenation/exercise.js
```diff
@@ -1,3 +1,5 @@
 // Start by creating a variable `message`
-
+let greeting = "My name is";
+let myName = "Boshra";
+let message = greeting + " " + myName;
```
Dedekind561, line 4: This looks good. How else do you think you can improve the RHS ( Right Hand Side ) on line 4 though? You're currently using `+` to concatenate the strings together - but what other techniques can you use for forming strings using variables. Try checking out [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#string_interpolation) and see if you can refactor your code :)

exercises/I-floats/exercise.js
```diff
@@ -1,2 +1,8 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+let totalNumber = numberOfMentors + numberOfStudents;
+let percentageOfStudents = (numberOfStudents / totalNumber) * 100;
+let roughPercentageOfStudents = Math.round(percentageOfStudents);
```
Dedekind561, line 5: Why did you decide to call this variable `roughPercentageOfStudents` in this case - what makes it rough?

exercises/K-functions-parameters/exercise3.js
```diff
@@ -1,5 +1,11 @@
 // Write your function here
+function createGreeting(string) {
+  return "my name is " + string;
+}
 
 var greeting = createGreeting("Daniel");
 
 console.log(greeting);
+
+var greeting = createGreeting("Boshra");
```
Dedekind561, line 10: :eyeglasses: Watch out for `var` here, again!

exercises/L-functions-nested/exercise.js
```diff
@@ -3,3 +3,17 @@ var mentor2 = "Irina";
 var mentor3 = "Mimi";
 var mentor4 = "Rob";
 var mentor5 = "Yohannes";
+
+function upperCaseName(mentor) {
+  return mentor.toUpperCase();
+}
+
+function greeting(mentor) {
+  return `HELLO ${upperCaseName(mentor)}`;
```
Dedekind561, line 12: Nice use of template literals here! :smile:

extra/2-piping.js
```diff
@@ -16,26 +16,30 @@
...
 
-// Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+// Why can this code be seen as bad practice? Comment your answer.//there is lots of function in one line which makes hard to understands for someone who read the codes
+let badCode = format(multiply(add(startingValue, 10), 2));
 
 /* BETTER PRACTICE */
-
-let goodCode =
+function giveNewValueTo(startingValue) {
```
Dedekind561, line 37: I think `startingValue` is an excellent variable name here. Nice.

mandatory/4-tax.js
```diff
@@ -17,8 +19,10 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
-
+function addTaxAndFormatCurrency(price) {
+  return `£${calculateSalesTax(price).toFixed(2)}`;
```
Dedekind561, line 23: Possibly think about adding another variable in here to make the final return value a bit more readable. 
Which part of the expression `calculateSalesTax(price).toFixed(2)` could you lift out of the template literal and give a meaningful name to ?

extra/1-currency-conversion.js
```diff
@@ -15,8 +17,10 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
-
+function convertToBRL(price) {
+  let newPrice = price * 0.99 * 5.7;
```
Dedekind561, line 21: This solution looks good to me - I like your use of multiplication by `0.99` to reduce the original value by 1%.
Think about this function for a moment from the perspective of somebody who perhaps is a bit less familiar.
They probably won't have much idea of what the `5.7` means.
Could you assign 5.7 to a variable with a clear name showing what it represents ?

Dedekind561 (commented):
Hi @BoshraM, 

Good job on completing your first piece of JS coursework at CYF! :tada: 

I'm seeing lots of good solutions in your coursework so far, so keep up the effort. I think my main piece of feedback would be to think about using some extra variables to store parts of the more complex expressions in your functions. Using variables like this will improve the readability of your code from day 1.

Nice work! :+1:

PR #427, opened 2022-11-22
London Class 9 - Abdulmajid Rammali - Javascript - Week 1

Abdulmajidram:

# Your Details

- Your Name: Abdulmajid Rammali
- Your City: London
- Your Slack Name: Abdulmajidram

# Homework Details

- Module:1
- Week:1

# Notes

- What did you find easy? exercise

- What did you find hard?
last few exercises
need revision on functions

- Any other notes?

### DIFF
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,3 +1,5 @@
-const name = " Daniel  ";
+const myName = " Daniel  ";
 
-console.log(message);
+
+console.log(`Hi I am ${myName.trim()} and my name has ${myName.length} letters`)
+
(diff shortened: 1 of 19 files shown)


exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,5 @@
-const name = " Daniel  ";
+const myName = " Daniel  ";
```
ZahraaTayyar, line 1: use "let"

ZahraaTayyar (commented):
left a little change. Well done!

exercises/F-strings-methods/exercise2.js
Abdulmajidram: Thanks for the feedback, is it for only this one or you mean in general? 
cause if in general then yeah I should of  used Ellie advice to use let :D

gladwindos (commented):
This looks really good! The only thing I would suggest is to get in the habit of using `const` instead of `let` or `var` when you know the variable you are creating is not going to change.

Abdulmajidram:
alright will do thanks for the feedback Gladwin.

PR #429, opened 2022-11-22
London 9 -Marziyeh Azhdari - JS-Core-1 - Week 1 

MariAzhdari:

# Your Details

- Your Name: Marziyeh Azhdari
- Your City:London
- Your Slack Name:MariAzh

# Homework Details

- Module:JS
- Week:1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -1,6 +1,6 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a, b ,c) {
   return a + b + c;
 }
 


extra/1-currency-conversion.js
```diff
@@ -5,7 +5,11 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(price) {
+  return price*1.4
```
GayleNThompsonIgwebike, line 9: I like this syntax...mine was sooo long... Great job.

GayleNThompsonIgwebike (commented):
Great job Marziyeh. Well done.

exercises/K-functions-parameters/exercise3.js
```diff
@@ -1,5 +1,7 @@
 // Write your function here
+function greeting(name){
+ console.log("Hello,my name is ;"+(name));
```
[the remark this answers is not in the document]
MariAzhdari: exactly .Thanks Gayle

exercises/B-hello-world/exercise.js
```diff
@@ -1 +1,6 @@
 console.log("Hello world");
+console.log("cd exercises/B-hello-world")
```
mcarballopacheco, line 2: Interesting test! I hope you realised that this will still just print "cd exercises/B-hello-world", even if this command has a very different meaning when you run it in the console, that is, this will not change directory when run inside the console.log as a string.

exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,5 @@
-const name = " Daniel  ";
-
-console.log(message);
+let myName="mari"
+let lengthName=myName.length
+// let messageTwo=`my name is${myName}and my name is${lengthName}characters long`
+let message="        My name is"+myName+" and my name is " + lengthName+" characters long"
```
mcarballopacheco, line 4: This is a very small comment, but you should try to be consistent with your style. If you are putting spaces before and after the plus signs, for example. `is" + myName + " an` you should do that for all strings concatenations.

exercises/I-floats/exercise.js
```diff
@@ -1,2 +1,7 @@
-var numberOfStudents = 15;
-var numberOfMentors = 8;
+let numberOfStudents = 15;
+let numberOfMentors = 8;
+let total=numberOfMentors+numberOfStudents
+let percentageOfStudent=numberOfStudents/total*100;
+console.log("Percentage students: "+Math.round(percentageOfStudent)+"%");
+let PercentageMentors=100-percentageOfStudent
```
mcarballopacheco, line 6: You could have also calculated the percentage of mentors, as.
`let PercentageMentors=numberOfMentors/total*100;`
and then use round. Surprisingly, this may not always be the same as the way you calculated it, because if the fraction ends in 0.5, it will always go up to the next integer when using Math.round. This is why in general, it is important to be aware and careful when rounding up numbers.

exercises/F-strings-methods/exercise2.js
MariAzhdari: I used the `.trim` method to remove the extra whitespace for this I made space. so I should do that for all strings concatenations??

exercises/K-functions-parameters/exercise3.js
```diff
@@ -1,5 +1,7 @@
 // Write your function here
+function greeting(name){
+ console.log("Hello,my name is ;"+ name);
```
mcarballopacheco, line 3: I think here it would have make more sense based on the exercise to return the greeting and then use console.log() outside of the function.

exercises/K-functions-parameters/exercise4.js
```diff
@@ -1,5 +1,7 @@
 // Declare your function first
-
+function sum(num1,num2){
+    return num1+num2
+}
 // Call the function and assign to a variable `sum`
 
-console.log(sum);
+console.log(sum(13,124));
```
mcarballopacheco, line 7: This will work, but in the exercise it mentioned to assign the value to a variable called sum, so:
`let sum = add(13, 124)
console.log(sum)`

exercises/I-floats/exercise.js
MariAzhdari: I got it thanks.

exercises/L-functions-nested/exercise.js
```diff
@@ -1,5 +1,18 @@
...
+function nameToUpperCase(name){
+    return name.toUpperCase();
+}
+function greeting(name){
+    return `Hello ${nameToUpperCase(name)}`;
+}
+function result(name){
+    return console.log(greeting(name));
+}
+result(mentor1);
```
mcarballopacheco, line 17: This looks great!

exercises/K-functions-parameters/exercise3.js
MariAzhdari: function greeting(name){
return name;
}
console.log("Hello,my name is ;"+greeting("Daniel"));


Is it ok?

extra/2-piping.js
```diff
@@ -16,21 +16,25 @@
...
+function format(number) {
+return `£${number}`
 }
+console.log(format(16))
 
 const startingValue = 2;
 
 // Why can this code be seen as bad practice? Comment your answer.
+//there are three functions that are not necessary and I think it is possible with a shorter code as well.(I am not sure!!!!!!!!!!!!!!!!!)
 let badCode =
```
mcarballopacheco, line 39: It is in general a bad idea to call many functions in the same line. It makes the code very hard to read. A better idea is to call each function once in a line and assign the result to a variable which then gets called in the following like but the next function, etc.

mandatory/3-function-output.js
```diff
@@ -1,17 +1,24 @@
 // Add comments to explain what this function does. You're meant to use Google!
+//The Math.random() function returns a floating-point, pseudo-random number that's greater than or equal to 0 and less than 1,
 function getRandomNumber() {
```
mcarballopacheco, line 3: I think here, the question is asking what does the `getRandomNumber()` does also?

exercises/F-strings-methods/exercise2.js
GayleNThompsonIgwebike: Thanks @mcarballopacheco Your suggestion is quite right. I'm practising consistency.

exercises/B-hello-world/exercise.js
GayleNThompsonIgwebike: I understand this. I think I was just messing about with random values to see my output when I run them and left this in... but thanks for the pick up and comment.

mandatory/4-tax.js
```diff
@@ -17,8 +21,13 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price) {
+   return `£${(price * 1.2).toFixed(2)}`;
```
mcarballopacheco, line 25: You should have used the calculateSalesTax function that you just created. The reason we create functions is to not repeat code which minimises errors. For example, let say the sales tax changes from 20% to 25%. if all of the times you calculated the sales tax you always use the calculateSalesTax function, then you only need to change the calculation in one place only. But if you are calculating in different places, then you would have to find them one by one.

mcarballopacheco (commented):
Amazing work Mari!

PR #436, opened 2022-11-24
update

mickeyhaile2:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?
getting a function into a function 

- Any other notes?


exercises/D-strings/exercise.js
```diff
@@ -1,3 +1,7 @@
 // Start by creating a variable `message`
+var message = "This is a string";
```
JDysiewicz, line 2: try to avoid using `var`; it can cause issues with scoping https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/ (you're not expected to understand what this means currently, just here in case you're interested). Try to always use `const`, then, if the variable needs reassigning, delcare with `let` instead

exercises/G-numbers/exercise.js
```diff
@@ -1 +1,6 @@
 // Start by creating a variables `numberOfStudents` and `numberOfMentors`
+numberOfStudents = 15;
+numberOfMentors = 8;
+total = numberOfMentors + numberOfStudents
```
JDysiewicz, line 4: Variables in JavaScript need to be "declared"; this means you prefix the variable name with `let` or `const` (e.g. `const numberOfStudents = x`, `let numberOfMentors = y`. This is effectively letting JS know that `numberOfStudents` refers to a variable, instead of being some other key word.

exercises/J-functions/exercise.js
```diff
@@ -1,7 +1,7 @@
 function halve(number) {
-  // complete the function here
+  return Number*.5
```
JDysiewicz, line 2: variables are case-sensitive; in this case you want to refer to the `number` variable; but instead you're referring to something called `Number` (which, incidentally, is a thing that exists in JS)

exercises/K-functions-parameters/exercise2.js
```diff
@@ -1,5 +1,8 @@
 // Declare your function first
-
+function name(a,b) {
+    return a/b
```
JDysiewicz, line 3: What would happen if `b = 0`; you would execute `a/0` - give this a try and see what happens; have a think about how you might handle that situation (note - this is more advanced than strictly necessary at this point, but it's always good to think ahead)

exercises/K-functions-parameters/exercise5.js
```diff
@@ -1,4 +1,7 @@
 // Declare your function here
+function createLongGreeting(a, b) {
+    return "I am " + a + " and I am " + b + " years old"
```
JDysiewicz, line 3: As a general comment, how might you rewrite this as a template string? Usually, template strings are preferred to string concatenation (especially when there are many variables involved), as it's usually more readable.

JDysiewicz (commented):
looking good so far, seems you're taking to JS well :D Careful of not using `var` (stick to `const`, or if _necessary_ use `let`), and remember that variables in JS are case-sensitive.

JDysiewicz:
Make sure to stick to the naming convention specified in the template for naming your PRs; it makes locating and organizing them far easier.

mandatory/4-tax.js
```diff
@@ -5,8 +5,13 @@
   Sales tax is 20% of the price of the product.
 */
 
+
 function calculateSalesTax() {}
 
+ function calculateSalesTax(sales) {
+   const calcTax = sales * 0.2 + sales;
+   return calcTax;
+ }
```
JDysiewicz, line 14: You're declaring the same function name twice here, which is why one of your tests is failing. You can remove the declaration on line 9.

PR #443, opened 2022-11-24
London Class 9 - Farzaneh Haghani - JavaScript - Week 1

farzaneh-haghani:
# Your Details

- Your Name: Farzaneh Haghani
- Your City: London
- Your Slack Name: Farzaneh Haghani

# Homework Details

- Module: JavaScript
- Week: 1

# Notes

- What did you find easy? Find syntax errors

- What did you find hard? 8 ball magic (It wasn't hard but it was unclear that how it works).

- What do you still not understand? Nothing

- Any other notes? No

### DIFF
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,3 +1,5 @@
 // Start by creating a variable `greeting`
-
+var greeting = "Hello world";
+console.log(greeting);
+console.log(greeting);
 console.log(greeting);
--- a/exercises/F-strings-methods/exercise.js
+++ b/exercises/F-strings-methods/exercise.js
@@ -1,3 +1,4 @@
 // Start by creating a variable `message`
-
-console.log(message);
+var myName = "Daniel";
+var message = myName.length;
+console.log(`My name is ${myName} and my name is ${message} characters long`);
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,3 +1,5 @@
 const name = " Daniel  ";
-
+message = name.trim(" ");
 console.log(message);
+console.log(`My name is ${message} and my name is ${message.length} characters long`);
+
--- a/exercises/K-functions-parameters/exercise2.js
+++ b/exercises/K-functions-parameters/exercise2.js
@@ -1,4 +1,7 @@
 // Declare your function first
+function divide(x, y) {
+    return x / y;
+}
 
 var result = divide(3, 4);
 
--- a/extra/2-piping.js
+++ b/extra/2-piping.js
@@ -16,30 +16,31 @@
   the final result to the variable goodCode
 */
 
-function add() {
-
+function add(firstNum, secondNum) {
+  return firstNum + secondNum;
 }
 
-function multiply() {
-
+function multiply(firstNum, secondNum) {
+  return firstNum * secondNum;
 }
 
-function format() {
-
+function format(num) {
+  return `£${num}`;
 }
 
 const startingValue = 2;
 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+let badCode = format(multiply(add(startingValue, 10), 2));
 
 /* BETTER PRACTICE */
-
-let goodCode =
+let result = add(startingValue, 10);
+result = multiply(result, 2);
+let goodCode = format(result);
 
 /* ======= TESTS - DO NOT MODIFY ===== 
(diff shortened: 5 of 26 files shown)


SaharNaderi (commented):
Thanks for hardworking .

exercises/C-variables/exercise.js
```diff
@@ -1,3 +1,5 @@
 // Start by creating a variable `greeting`
-
+var greeting = "Hello world";
```
JDysiewicz, line 2: try to always use `const` to declare variables; `let` can be used too but only use if if the variable _needs_ redefining at some point (normally, this won't be the case, which is why we prefer `const`).

`var` is now outdated, so try to avoid using it :P

exercises/F-strings-methods/exercise.js
```diff
@@ -1,3 +1,4 @@
 // Start by creating a variable `message`
-
-console.log(message);
+var myName = "Daniel";
+var message = myName.length;
+console.log(`My name is ${myName} and my name is ${message} characters long`);
```
JDysiewicz, line 4: nice use of a template string!

exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,5 @@
 const name = " Daniel  ";
-
+message = name.trim(" ");
```
JDysiewicz, line 2: almost - variables in JS need *declaring* by prefixing the variable name with `let` or `const`. This effectively tells JS "Hey, every time you see `variableName` from this point onwards, it's actually talking about this variable"

exercises/K-functions-parameters/exercise2.js
```diff
@@ -1,4 +1,7 @@
 // Declare your function first
+function divide(x, y) {
+    return x / y;
+}
```
JDysiewicz, line 4: You didn't have to account for this, but what would happen if `y = 0` was passed in and you end up with `x / 0`? try doing that in JS and see what happens. With knowledge of conditionals (like `if` statements) how would you prevent that from occurring?

extra/2-piping.js
```diff
@@ -16,30 +16,31 @@
...
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+let badCode = format(multiply(add(startingValue, 10), 2));
 
 /* BETTER PRACTICE */
-
-let goodCode =
+let result = add(startingValue, 10);
+result = multiply(result, 2);
+let goodCode = format(result);
```
JDysiewicz, line 39: Yup this is certainly better; one more  improvement that could be made here though is to use `const` instead to declare multiple variables e.g.
```
const addResult = add(startingValue, 10)
const multiplyResult = multiply(addResult, 2)
const formattedResult = format(multiplyResult)
```
A nice rule of thumb to go by with code-style is to always err on the more verbose side (i.e. it's almost always better to be more explicit about your intentions). The code will give the same result either way, so all your code-style is for is for other developers who'll look at your code, and in that case it's better to make your intentions very obvious.

extra/3-magic-8-ball.js
```diff
@@ -45,8 +45,41 @@
...
+  "My sources say no",
+  "Outlook not so good",
+  "Very doubtful"]];
+
+
 function shakeBall() {
   //Write your code in here
+
+  let firstItem = Math.floor(Math.random() * 4);
+  let secondItem = Math.floor(Math.random() * 5);
```
JDysiewicz, line 79: So `Math.random()` returns a random number _between 0 and 1_ (e.g. `0.6291232`). Therefore, when trying to get random _integers_ (i.e. for indexing an array) we tend to multiply by *array.length*, then get the `floor`. This means that we generate a random number between 0 and `array.length`, then get the `floor` (round down to nearest integer). This way, we generate a random number between `0 - array.length`, then get the `floor`, such that we are guaranteed a random number that is either 0,1,2,3...array.length-1 (remember, the last index of an array is it's length -1 as we start at `0`).

You _could_ avoid hard coding the 4 or 5 here by getting the array lengths, but that's beyond what is expected here

JDysiewicz (commented):
Looks good, well done! Hopefully the comment about 8 ball cleared things up a little.

exercises/K-functions-parameters/exercise2.js
farzaneh-haghani: Thanks. I updated

extra/3-magic-8-ball.js
farzaneh-haghani: Thanks. updated.

PR #447, opened 2022-11-25
London9 - Lorena Capraru - Javascript- Week2

LorenaCapraru:

# Your Details

- Your Name: Lorena
- Your City: London
- Your Slack Name: Lorena

# Homework Details

- Module: Javascript 
- Week: 2

# Notes

- What did you find easy?
- everything

- What did you find hard?
- nothing

- i understood it all.

- Any other notes?

### DIFF
--- a/exercises/F-strings-methods/exercise2.js
+++ b/exercises/F-strings-methods/exercise2.js
@@ -1,3 +1,3 @@
-const name = " Daniel  ";
-
+const nameD = " Daniel  ";
+let message ="My name is " +nameD.trim() +" and my name is " + nameD.trim().length +" characters long.";
 console.log(message);
--- a/exercises/I-floats/exercise.js
+++ b/exercises/I-floats/exercise.js
@@ -1,2 +1,9 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+var sum = numberOfMentors + numberOfStudents;
+
+var studentsPercentage = (numberOfStudents / sum) * 100;
+var mentorsPercentage = (numberOfMentors / sum) * 100;
+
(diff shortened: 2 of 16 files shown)


ehwus (commented):
This is an excellent submission - the tests are passing, and the code is readable and makes sense. I would just:
- Ensure that you're using `var` / `let` / `const` consistently in your code (https://www.freecodecamp.org/news/var-let-and-const-whats-the-difference/)
- Make sure that your variable names are descriptive: `fName` for instance tells me less than `firstName`. As long as you aren't excessively verbose, it's better to give more detail than leave ambiguity in naming.

Great effort, looking forward to seeing your CS skills continue to come through like this throughout the course!

exercises/I-floats/exercise.js
```diff
@@ -1,2 +1,9 @@
 var numberOfStudents = 15;
 var numberOfMentors = 8;
+var sum = numberOfMentors + numberOfStudents;
```
SagharHosseinmardi, line 3: Hi as a new developer it is better to use **let** or **const** instead of using **var.** The fact about const and let is you can’t re-assign another value to the constant x, however let allows you to declare variables.

exercises/F-strings-methods/exercise2.js
```diff
@@ -1,3 +1,4 @@
 const nameD = " Daniel  ";
 let message ="My name is " +nameD.trim() +" and my name is " + nameD.trim().length +" characters long.";
+//print the message
```
SagharHosseinmardi, line 3: Hi as a new developer it is better to use let or const instead of using var. The fact about const and let is you can’t re-assign another value to the constant x, however let allows you to declare variables.

PR #451, opened 2022-11-25
London 9 - Karleen Richards - Javascript - Core 1 - Coursework - Week 1

karleenmsrichards:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/2-piping.js
+++ b/extra/2-piping.js
@@ -16,26 +16,41 @@
   the final result to the variable goodCode
 */
 
-function add() {
-
+function add(a,b) {
+  return +(a + b).toFixed(1);
 }
 
-function multiply() {
+console.log(add(1,3));
+console.log(add(2.4,5))
 
+function multiply(c,d) {
+  return c * d;
 }
 
-function format() {
+console.log(multiply(2,3))
 
+function format(str) {
+  return "£" + str;
 }
 
-const startingValue = 2;
+console.log(format(16));
+console.log(format(10.1));
 
+const startingValue = 2; 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+//It is bad practice because the code is difficult to understand and read.
+let badCode = format(multiply(add(startingValue,10),2));
+
+console.log(badCode);
+
+
 
 /* BETTER PRACTICE */
+let first = startingValue + 10;
+let second = first * 2;
+let goodCode = format(second);
 
-let goodCode =
+console.log(goodCode);
 
 /* ======= TESTS - DO NOT MODIFY ===== 
 There are some Tests in this file that will help you work out if your code is working.
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -1,16 +1,19 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a,b,c) {
   return a + b + c;
 }
 
-function introduceMe(name, age)
-  return "Hello, my name is " + name "and I am " age + "years old";
+function introduceMe(name, age) {
+  name = "Sonjide";
+  age = 27;
+  return "Hello, my name is " + name + " and I am " + age + " years old";
+}
 
 function getTotal(a, b) {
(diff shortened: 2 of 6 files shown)


extra/2-piping.js
```diff
@@ -16,26 +16,41 @@
   the final result to the variable goodCode
 */
 
-function add() {
-
+function add(a,b) {
+  return +(a + b).toFixed(1);
```
Meysam-Arshadi, line 20: Great job! 
I think the task was to give the total No 2 decimals .toFixed(2)

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,19 @@
...
+}
 
 function getTotal(a, b) {
-  total = a ++ b;
-
-  return "The total is total";
+  a = 23;
+  b = 5;
+  return "The total is " + (a + b);
 }
```
PeteLindsell, line 17: This wants to be something like the following:

```
function getTotal(a, b) {
  const total = a + b;
  return "The total is total " + total;
}
```

By reassigning a and b in the function means you are throwing away the values being passed in

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,19 @@
...
   return a + b + c;
 }
 
-function introduceMe(name, age)
-  return "Hello, my name is " + name "and I am " age + "years old";
+function introduceMe(name, age) {
+  name = "Sonjide";
+  age = 27;
+  return "Hello, my name is " + name + " and I am " + age + " years old";
+}
```
PeteLindsell, line 11: the return string is correct but you dont want to reassign `name` and `age` to static values otherwise this function cant be used fore anyone that is not called Sonjide and happens to be 27 years old

extra/2-piping.js
```diff
@@ -16,26 +16,41 @@
...
+let badCode = format(multiply(add(startingValue,10),2));
+
+console.log(badCode);
+
+
 
 /* BETTER PRACTICE */
+let first = startingValue + 10;
+let second = first * 2;
+let goodCode = format(second);
```
PeteLindsell, line 51: Well done on getting onto this!

mandatory/4-tax.js
```diff
@@ -17,7 +22,12 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(b) {
+  return "£" + ((20 / 100 * b) + b).toFixed(2);
+}
```
PeteLindsell, line 27: Good use of .toFixed()

`(20 / 100 * b) + b` can be written as `b * 1.2` but your version is better when the percentage is being passed in as a value eg 20

PeteLindsell (commented):
Well done on getting through the mandatory tasks and onto some of the extra tasks!

I think you may have missed the exercises section.

mandatory/4-tax.js
karleenmsrichards: Hi Pete, I'm not sure if I understand what you mean. Can you please explain this more. Should I change it to b * 1.2?

extra/2-piping.js
karleenmsrichards: Hi hi, thank you. But my interpretation of this test: "test("add function - case 2 works", () => {
  expect(add(2.4, 5)).toEqual(7.4);
})"  is that I should ensure that there is one decimal place and not 2 for example if I use .toFixed(2) as you suggested, my answer may end up like 7.40 instead of 7.4 as an example left in the test. 

I could be wrong but that is how I looked at it. 

Also if I did not put .toFixed(1) I would get the recommended 7.4  and 10.1 but again I am not too sure what they expected hence I added .toFixed(1)

karleenmsrichards:
Thank you Pete. I have made the necessary adjustments apart from the one that I need clarification on. Also I did the exercises since September they are probably in another repo. I will sort it out and get back to you. Thanks much.

PR #453, opened 2022-11-25
Js week1

ebrahimbeiati:
Doing Js week 1 exercise

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/exercises/C-variables/exercise.js
+++ b/exercises/C-variables/exercise.js
@@ -1,3 +1,4 @@
 // Start by creating a variable `greeting`
-
-console.log(greeting);
+let greeting = new greeting;
+let greeting1 ="hin";
+console.log(greeting)
--- a/mandatory/2-logic-error.js
+++ b/mandatory/2-logic-error.js
@@ -1,16 +1,16 @@
 // The syntax for this function is valid but it has an error, find it and fix it.
 
-function trimWord(word) {
-  return wordtrim();
+function trimWord(word){
+  return word.trim();
 }
 
-function getStringLength(word) {
-  return "word".length();
+function getStringLength(word){
+  return word.length;
 }
 
-function multiply(a, b, c) {
-  a * b * c;
-  return;
+function multiply(a, b, c){
+    return a * b * c;
+
 }
 
 /* 
@@ -25,11 +25,11 @@ To run the tests for just this one file, type `npm test -- --testPathPattern 2-l
 */
 
 test("trimWord trims leading and trailing whitespace", () => {
-  expect(trimWord("   CodeYourFuture ")).toEqual("CodeYourFuture");
+  expect(trimWord("CodeYourFuture")).toEqual("CodeYourFuture");
 });
 
 test("trimWord doesn't remove whitespace in the middle of the string", () => {
-  expect(trimWord(" CodeYourFuture teaches coding     ")).toEqual(
+  expect(trimWord("CodeYourFuture teaches coding")).toEqual(
     "CodeYourFuture teaches coding"
   );
 });
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -2,13 +2,16 @@
 function getRandomNumber() {
   return Math.random() * 10;
 }
+//The Math.random() function returns a floating-point, pseudo-random number that's greater than or equal to 0 and less than 1, with approximately uniform distribution over that range — which you can then scale to your desired range. 
 
 // Add comments to explain what this function does. You're meant to use Google!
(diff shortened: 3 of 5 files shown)


mandatory/4-tax.js
```diff
@@ -5,7 +5,11 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  let salesTax = price * 0.2;
```
karleenmsrichards, line 9: Hello, I believe an error was made on this line. * means to multiply. I believe yous should divide price by 0.2 and not multiply.

mandatory/4-tax.js
```diff
@@ -17,8 +21,11 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price) {
+  return "£" + calculateSalesTax(price).toFixed(2);
```
karleenmsrichards, line 25: This function was written correctly but will not pass the test because it uses "calculateSalesTax" which had an error.

JDysiewicz:
Please use the naming format for PRs which is specified in the template; it makes it far easier to track these in the future

mandatory/2-logic-error.js
```diff
@@ -25,11 +25,11 @@ To run the tests for just this one file, type `npm test -- --testPathPattern 2-l
 */
 
 test("trimWord trims leading and trailing whitespace", () => {
-  expect(trimWord("   CodeYourFuture ")).toEqual("CodeYourFuture");
+  expect(trimWord("CodeYourFuture")).toEqual("CodeYourFuture");
 });
 
 test("trimWord doesn't remove whitespace in the middle of the string", () => {
-  expect(trimWord(" CodeYourFuture teaches coding     ")).toEqual(
+  expect(trimWord("CodeYourFuture teaches coding")).toEqual(
```
JDysiewicz, line 32: The good news is your `trimWord` function looks perfect :D However, please **never** change the tests in these files. They are there to indicate if your code is working as it should, and by changing them you are opening yourself to a world of issues. If you think there _is_ a problem with one of the tests, please let myself (or another volunteer) know so we can investigate, but please **don't change them yourself**.

exercises/C-variables/exercise.js
```diff
@@ -1,3 +1,4 @@
 // Start by creating a variable `greeting`
-
-console.log(greeting);
+let greeting = new greeting;
+let greeting1 ="hin";
```
JDysiewicz, line 3: good to see you using `let` instead of `var`, a small improvement would be to use `const` instead (as you never redefine the variable). Also, you don't need to use the `new` keyword here (that is used for other things we'll discuss at a later point). Here, just the `let greeting = 'hin'` would be fine.

mandatory/3-function-output.js
```diff
@@ -2,13 +2,16 @@
...
+//The Math.random() function returns a floating-point, pseudo-random number that's greater than or equal to 0 and less than 1, with approximately uniform distribution over that range — which you can then scale to your desired range. 
 
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
+//The concat() method joins two or more strings.
 
 function concatenate(firstWord, secondWord, thirdWord) {
+  return `${firstWord} ${secondWord} ${thirdWord}`;
```
JDysiewicz, line 14: nice use of a template string!

mandatory/4-tax.js
```diff
@@ -28,7 +35,7 @@ There are some Tests in this file that will help you work out if your code is wo
 To run the tests for just this one file, type `npm test -- --testPathPattern 4-tax` into your terminal
 (Reminder: You must have run `npm install` one time before this will work!)
 ===================================================
-*/
+// */
```
JDysiewicz, line 38: again, please don't modify anything in the test section (see the line which states "TESTS - DO NOT MODIFY BELOW THIS LINE")

JDysiewicz (commented):
Overall it's looking pretty good, you seem to be getting to grips with JS pretty well. Just be careful not to modify tests in future PRs

JDysiewicz:
It also looks like you've only done `exercise/C-variables` - make sure to do the others when you have time (you have to do all the things in `/exercises` _and_ `/mandatory`).

PR #461, opened 2022-12-01
fix: replace var with const keyword

Dedekind561:


SallyMcGrath (approved):
LGTM

I feel physically ill approving a PR with failing tests but I do see there's no way round this. 

Let @moneyinthesky know you have made this change as I do recall he raised a PR on the solutions repo for this week.

PR #473, opened 2023-02-16
Glasgow Class 6 - Malkit Benning - JS - Week 1

malkitbenning:

# Your Details

- Your Name: Malkit Benning
- Your City: Glasgow
- Your Slack Name: Malkit Benning

# Homework Details

- Module: JS
- Week: 1

# Notes

- What did you find easy? Video explanations

- What did you find hard? Jest in general. Coding from a test perspective.

- What do you still not understand? Where I went wrong with last answer for Magic 8 Ball

- Any other notes?

### DIFF
--- a/extra/2-piping.js
+++ b/extra/2-piping.js
@@ -16,26 +16,29 @@
   the final result to the variable goodCode
 */
 
-function add() {
-
+function add(a, b) {
+  return a + b;
 }
 
-function multiply() {
-
+function multiply(a, b) {
+  return a * b;
 }
 
-function format() {
-
+function format(inFigure) {
+  return `£${inFigure}`;
 }
 
 const startingValue = 2;
 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+// confusing to understand with the nested commands
+let badCode = format(multiply(add(startingValue, 10), 2));
 
 /* BETTER PRACTICE */
 
-let goodCode =
+let goodCode = add(startingValue, 10);
+goodCode = multiply(goodCode, 2);
+goodCode = format(goodCode);
 
 /* ======= TESTS - DO NOT MODIFY ===== 
 There are some Tests in this file that will help you work out if your code is working.
--- a/extra/3-magic-8-ball.js
+++ b/extra/3-magic-8-ball.js
@@ -71,7 +71,6 @@ function shakeBall() {
     "Very doubtful."
   ];
   let randomNum = Math.floor(Math.random() * 20) + 1;
-
   return predictions[randomNum];
 }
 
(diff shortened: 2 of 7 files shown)


extra/3-magic-8-ball.js
```diff
@@ -60,6 +85,69 @@ function shakeBall() {
 */
 function checkAnswer(answer) {
   //Write your code in here
+
```
Mpanasetckiy, line 88: Veryyyy long if else  loop don't you think? Try to break down your array into 4 small ones. Then we can use them to compare if the inbound answer is inside one of the arrays. Also, this function doesn't correlate with your array at all.

extra/3-magic-8-ball.js
```diff
@@ -47,6 +47,31 @@
 // and return the answer.
 function shakeBall() {
   //Write your code in here
+  console.log("The ball has shaken!");
+  const predictions = [
```
Mpanasetckiy, line 51: We also can set our array outside of the function, for better readability.

extra/3-magic-8-ball.js
malkitbenning: Wow! Maksim - this is so useful. Thank you for taking the time to review! I'm also very impressed that you fixed the bug in your version. I made a few corrections based on your advice. Hope I understood your feedback correctly.

mandatory/4-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(preTax) {
+  return preTax * 1.2;
```
Melesegithub, line 9: Hi Malkit, You doing great even you cover the extra one. in calculateSalesTax function return you multiply preTax with 1.2. I think it should be 0.2.

mandatory/3-function-output.js
```diff
@@ -1,14 +1,17 @@
 // Add comments to explain what this function does. You're meant to use Google!
+// This function generates a random number between 0 and 10
```
msimmdev, line 2: Thought: This is correct, but it's worth noting that the number may not be an integer, and could be 0 but cannot be 10.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random for more info

mandatory/3-function-output.js
```diff
@@ -1,14 +1,17 @@
...
 }
 
 // Add comments to explain what this function does. You're meant to use Google!
+// The concat() method joins two or more strings.
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
 
 function concatenate(firstWord, secondWord, thirdWord) {
+  return firstWord.concat(" ", secondWord, " ", thirdWord);
```
msimmdev, line 14: Nitpick: This is fine, but reads a little funny to me. you could also do something like this
``
return `${firstWord} ${secondWord} ${thirdWord}`
``

msimmdev (commented):
Praise: Generally very good answers to the exercises Mal. Well done on doing the extra exercises as well.

mandatory/4-tax.js
malkitbenning: I agree Michael then name of the function could be more meaningful.

extra/2-piping.js
```diff
@@ -16,26 +16,29 @@
...
 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+// confusing to understand with the nested commands
+let badCode = format(multiply(add(startingValue, 10), 2));
 
 /* BETTER PRACTICE */
 
-let goodCode =
+let goodCode = add(startingValue, 10);
```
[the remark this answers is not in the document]
malkitbenning: Thanks Michael, I changed the names of the intermediate values, but kept the last variable name as goodCode as it is used in the actual test.

PR #476, opened 2023-02-17
do jest-extended setup in each project

### LINKED ISSUE #224: toBeOneOf in extra 3-magic-8-ball test not working
Screenshot 2021-12-12 at 15 53 02
Screenshot 2021-12-12 at 15 52 53
I've found a problem with the test toBeOneOf I've added see the screenshot

Dedekind561:
Fixes #224 

I think the issue here is the jest-extended setup is not being applied to each project defined inside the `package.json`. Wondering if there is a nicer way of doing this but this fix works for the time being.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,21 +14,30 @@
     "url": "https://github.com/CodeYourFuture/JavaScript-Core-1-Coursework-Week1/issues"
   },
   "jest": {
-    "setupFilesAfterEnv": ["jest-extended"],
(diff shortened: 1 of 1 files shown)


Mpanasetckiy:
Thanks, it worked.

gregdyke (approved):

gregdyke:
I think we expect the tests to fail here, so I've overridden and merged

PR #482, opened 2023-02-18
 Glasgow 6 - Mandy Cheung - JS 1 - Week 5?

m4ndycheung:

# Your Details

- Your Name: Mandy Cheung
- Your City: Glasgow
- Your Slack Name: Mandy Cheung

# Homework Details

- Module: JS 1
- Week: 5

# Notes

- What did you find easy? 
- Finding the syntax errors

- What did you find hard?
- getting my head around functions - understanding inputs

- Any other notes?

### DIFF
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -1,16 +1,17 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a, b, c) {
   return a + b + c;
 }
 
-function introduceMe(name, age)
-  return `Hello, my {name}` is "and I am $age years old`;
+function introduceMe(name, age) {
+  return `Hello, my name is ${name} and I am ${age} years old`;
+}
 
 function getTotal(a, b) {
-  total = a ++ b;
+  total = a + b;
 
-  return "The total is total";
+  return `The total is ${total}`;
 }
 
 /* 
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -3,15 +3,24 @@ function getRandomNumber() {
   return Math.random() * 10;
 }
 
+// This function will return a random number greater than or equal to 1 or less than 10. It will be a decimal. We don't need to pass an argument through this function.
+
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
 
+// This function will combine 2 or more arguments passed through this function.
+// In this function, word1 will be combined with word2 in that order as word2 is inside the brackets.
+// This function will create a new array, it does not change word1 or word2.
+
 function concatenate(firstWord, secondWord, thirdWord) {
+  return firstWord.concat(" ", secondWord, " ", thirdWord);
   // Write the body of this function to concatenate three words together.
   // Look at the test case below to understand what this function is expected to return.
 }
+
+// Without the " " the words concatenate without spaces between the words. it returned some wild words. I added " " to the concat function to add spaces.
 
 /* 
 ===================================================
--- a/mandatory/4-tax.js
+++ b/mandatory/4-tax.js
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(priceNoTax) {
+  return priceNoTax + (priceNoTax * 0.2);
+}
 
 /*
   CURRENCY FORMATTING
@@ -17,7 +19,9 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(priceNoTax) {
+  return "£" + (priceNoTax + (priceNoTax * 0.2)).toFixed(2);
+}
 
 /* 
 ===================================================
--- a/mandatory/2-logic-error.js
+++ b/mandatory/2-logic-error.js
@@ -1,16 +1,15 @@
 // The syntax for these functions is valid but there are some errors, find them and fix them
 
 function trimWord(word) {
-  return wordtrim();
+  return word.trim();
 }
 
 function getStringLength(word) {
-  return "word".length();
+  return word.length;
 }
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+  return a * b * c;
 }
 
 /* 


mandatory/3-function-output.js
```diff
@@ -3,16 +3,25 @@ function getRandomNumber() {
...
+// This function will return a random number greater than or equal to 1 or less than 10. It will be a decimal. We don't need to pass an argument through this function.
+
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
 
+// This function will combine 2 or more arguments passed through this function.
+// In this function, word1 will be combined with word2 in that order as word2 is inside the brackets.
+// This function will create a new array, it does not change word1 or word2.
```
lazar-eva, line 15: Hey Mandy, I think you've done a great job in solving all the challenges, nice and clean code. The only note I have - concat method returns a string not an array :)

m4ndycheung: Thanks for reviewing my code, Alena! 
I must've taken that bit about arrays from my google search it says: 
"The concat() method is used to merge two or more arrays. This method does not change the existing arrays, but instead returns a new array. "

In this exercise we're not working with arrays so you're right here - thanks for pointing that out :)

extra/2-piping.js
```diff
@@ -16,26 +16,35 @@
...
 const startingValue = 2;
 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+
+// So, I'm not too sure what bad code or good code looks like yet so I wrote this, it seems vague and repetitive:
+// let badCode = add(10, startingValue); badCode = multiply(badCode, 2); badCode = format(badCode);
+
+// Below, I wrote some more of what I interpret bad code to look like too. This way seems complicated for me to read functions nested in others (although not sure if that's down to me being a beginner!):
+let badCode = format(multiply(add(10, startingValue), 2));
```
PaulinaWywrot, line 39: Hi Mandy, I think you did a great job, exactly what was supposed to be done in this exercise. 
It's definitely not you being a beginner, I don't think it would be easy to understand this code for anybody, no matter what level of knowledge they had.

PaulinaWywrot (commented):
Good job Mandy, I think your code is very organised and neatly written. I also like your notes added to some of the exercises.

mandatory/3-function-output.js
m4ndycheung: @lazar-eva My mistake, I read the wrong mdn document, I hadn't realised there are two ways to use concat:

1. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat
2. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat

I'll change my answer, thanks again!

extra/2-piping.js
m4ndycheung: Hi Paulina, thank you for reviewing my code and for your reassuring comments :)

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,17 @@
...
 
-function introduceMe(name, age)
-  return `Hello, my {name}` is "and I am $age years old`;
+function introduceMe(name, age) {
+  return `Hello, my name is ${name} and I am ${age} years old`;
+}
 
 function getTotal(a, b) {
-  total = a ++ b;
+  total = a + b;
```
msimmdev, line 12: Suggestion: use the let keyword to declare the variable

e.g. `let total = a + b`

The code will work without it but using let scopes the variable to the getTotal function. Without let is there was another function that used a variable called total you could see conflicts.

mandatory/3-function-output.js
```diff
@@ -3,16 +3,24 @@ function getRandomNumber() {
   return Math.random() * 10;
 }
 
+// This function will return a random number greater than or equal to 1 or less than 10. It will be a decimal. We don't need to pass an argument through this function.
```
msimmdev, line 6: Issue: This is not quite correct, the number will be greater than or equal to 0 and less than 10
For more info see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random

mandatory/3-function-output.js
```diff
@@ -3,16 +3,24 @@ function getRandomNumber() {
...
+
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
 
+// This function will combine 2 or more arguments passed through this function.
```
msimmdev, line 13: Nitpick: This function will combine exactly 2 arguments, no more

mandatory/3-function-output.js
```diff
@@ -3,16 +3,24 @@ function getRandomNumber() {
...
+// In this function, word1 will be combined with word2 in that order as word2 is inside the brackets. The function will return a new string.
+
 function concatenate(firstWord, secondWord, thirdWord) {
+  return firstWord.concat(" ", secondWord, " ", thirdWord);
```
msimmdev, line 17: Nitpick: This is fine, but reads a little funny to me. you could also do something like this
``
return `${firstWord} ${secondWord} ${thirdWord}`
``

mandatory/4-tax.js
```diff
@@ -5,7 +5,9 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(priceNoTax) {
```
msimmdev, line 8: Thought: What you've done is exactly right, however I think the function name you've been given is misleading as it suggests it will return the sales tax only and not the price with sales tax added. Maybe a better name would be `calculateAndAddSalesTax`

mandatory/4-tax.js
```diff
@@ -17,7 +19,9 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(priceNoTax) {
+  return "£" + (priceNoTax + (priceNoTax * 0.2)).toFixed(2);
```
msimmdev, line 23: Suggestion: Rather than re-implementing the sales tax logic in this function it would make sense to call the `calculateSalesTax` function you wrote above

extra/1-currency-conversion.js
```diff
@@ -15,7 +17,9 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(amountGBP) {
+  return Number((0.99 * amountGBP * 5.7).toFixed(2));
```
msimmdev, line 21: Suggestion: Avoid using the toFixed for math operations, as toFixed converts the number to a string. For a pure math function it is better to use the `Math.round(val)` function to avoid the conversion to a string.

extra/3-magic-8-ball.js
```diff
@@ -43,10 +43,48 @@
...
+]
+
+let allAnswers = veryPositiveAnswers.concat(positiveAnswers, negativeAnswers, veryNegativeAnswers);
+
 // This should log "The ball has shaken!"
 // and return the answer.
 function shakeBall() {
-  //Write your code in here
+  console.log(`The ball has shaken!`);
+  randomAnswer = allAnswers[Math.floor(Math.random() * allAnswers.length)];
```
msimmdev, line 86: Nitpick: it would be better to use `let` to declare the randomAnswer variable here in the function rather than further up in the script.

msimmdev (commented):
Praise: Really good work Mandy, well done on completing the extra questions I found no issues in those answers

mandatory/1-syntax-errors.js
m4ndycheung: Thanks for the feedback, Michael. So if I don't want total to be used outside of the getTotal function, I need to declare total using let.

mandatory/3-function-output.js
m4ndycheung: I hadn't thought of that. Thanks, Michael.

extra/3-magic-8-ball.js
m4ndycheung: That's good to know, thanks!

PR #501, opened 2023-02-19
Glasgow-6-Haroun- JavaScript-Core-1-Coursework-Week1

HarounAlarabi:
- Your Name:Haroun
- Your City:Glasgow
- Your Slack Name:HarounAlarabi

# Homework Details

- Module:javascript core 1
- Week:1

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?


Muath-Alawadhi:
Well done ^ ^

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,19 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
-  return a + b + c;
+function addNumbers(a, b ,c) {
+  let d = a + b + c;
```
Abdirizak-Mohamed, line 4: Couple thoughts, do we need  a variable here? And if we do can we name it better?

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,19 @@
...
+  return `Hello, my name is ${name} and I am ${age} years old`;
+}
+ 
 
 function getTotal(a, b) {
-  total = a ++ b;
+  total = a + b;
 
-  return "The total is total";
+  return "The total is"+" "+ `${total}`;
```
Abdirizak-Mohamed, line 16: Question: So here we are using [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) which is nice!! But if so do we need to add strings together? i.e. can we 

```
const personName = "Abdi"

`The greatest name in the world is ${personName}` 
```

mandatory/2-logic-error.js
```diff
@@ -1,16 +1,16 @@
...
-  return "word".length();
+  return word.length;
 }
 
 function multiply(a, b, c) {
-  a * b * c;
-  return;
+   let x = a * b * c;
+  return x;
 }
```
Abdirizak-Mohamed, line 15: Praise : Noice 👍

mandatory/3-function-output.js
```diff
@@ -2,13 +2,18 @@
...
+//This function is returning a random number between 0  and 10 
 
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
+//The concat() method joins two or more strings together into a new string without modifying the original strings
 
 function concatenate(firstWord, secondWord, thirdWord) {
+  return firstWord + " "+ secondWord + " "+thirdWord;
```
Abdirizak-Mohamed, line 14: Question: Have we answered the question here? I think the goal is to use the concat string method here. 

i.e. 

```
const wordOne = "Hello"
const wordTwo = "World"

const combinedWord = wordOne.concat(wordTwo)
```

How would you concat three words like above?

Abdirizak-Mohamed (approved):

PR #510, opened 2023-02-21
LONDON_10 | SAIM KORKMAZ | JS-1-1

nsaimk:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/extra/1-currency-conversion.js
+++ b/extra/1-currency-conversion.js
@@ -5,7 +5,11 @@
   Write a function that converts a price to USD (exchange rate is 1.4 $ to £)
 */
 
-function convertToUSD() {}
+function convertToUSD(price) {
+  return price * 1.4;
+}
+
+
 
 /*
   CURRENCY CONVERSION
@@ -15,7 +19,10 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(price) {
+  price = price * 5.7 * 0.99;
+  return parseFloat(price.toFixed(2));
+}
 
 /* ======= TESTS - DO NOT MODIFY ===== 
 There are some Tests in this file that will help you work out if your code is working.
--- a/extra/2-piping.js
+++ b/extra/2-piping.js
@@ -16,26 +16,27 @@
   the final result to the variable goodCode
 */
 
-function add() {
-
+function add(x, y) {
+  return x + y;
 }
 
-function multiply() {
-
+function multiply(x, y) {
+  return x * y;
 }
 
-function format() {
-
+function format(x) {
+  return `£${x}`;
 }
 
 const startingValue = 2;
 
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+let badCode = format(multiply(add(startingValue, 10), 2));
 
 /* BETTER PRACTICE */
-
-let goodCode =
+let step1 = add(startingValue, 10);
+let step2 = multiply(step1, 2);
+let goodCode = format(step2);
 
 /* ======= TESTS - DO NOT MODIFY ===== 
 There are some Tests in this file that will help you work out if your code is working.
--- a/extra/2-piping.js
+++ b/extra/2-piping.js
@@ -34,9 +34,9 @@ const startingValue = 2;
 let badCode = format(multiply(add(startingValue, 10), 2));
 
 /* BETTER PRACTICE */
-let step1 = add(startingValue, 10);
-let step2 = multiply(step1, 2);
-let goodCode = format(step2);
+const step1 = add(startingValue, 10);
+const step2 = multiply(step1, 2);
+const goodCode = format(step2);
 
 /* ======= TESTS - DO NOT MODIFY ===== 
 There are some Tests in this file that will help you work out if your code is working.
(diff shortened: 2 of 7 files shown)


extra/1-currency-conversion.js
```diff
@@ -15,7 +19,10 @@ function convertToUSD() {}
   They have also decided that they should add a 1% fee to all foreign transactions, which means you only convert 99% of the £ to BRL.
 */
 
-function convertToBRL() {}
+function convertToBRL(price) {
+  price = price * 5.7 * 0.99;
+  return parseFloat(price.toFixed(2));
+}
```
berkeli, line 25: Please look into pure functions and argument mutation. 

It's considered a bad practice for a function to have side effects. Can you think of which side effect this function has and why would it be bad practice?

extra/2-piping.js
```diff
@@ -16,26 +16,27 @@
...
 // Why can this code be seen as bad practice? Comment your answer.
-let badCode =
+let badCode = format(multiply(add(startingValue, 10), 2));
 
 /* BETTER PRACTICE */
-
-let goodCode =
+let step1 = add(startingValue, 10);
+let step2 = multiply(step1, 2);
+let goodCode = format(step2);
```
berkeli, line 39: We should try to use `const` if we know the value is not going to be re-assinged/changed.

extra/3-magic-8-ball.js
```diff
@@ -60,6 +85,34 @@ function shakeBall() {
...
+    answer == "Reply hazy, try again." ||
+    answer == "Ask again later." ||
+    answer == "Better not tell you now." ||
+    answer == "Cannot predict now." ||
+    answer == "Concentrate and ask again."
+  ) {
+    return "negative";
+  } else {
+    return "very negative";
+  }
```
berkeli, line 115: This is fine, but you are using a lot of magic words here. Try to look into what magic words are and how this could be improved.

Hint: 
You can either use a JS object or array index to determine whether it's positive or negative

mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,16 @@
...
 
-function introduceMe(name, age)
-  return `Hello, my {name}` is "and I am $age years old`;
-
+function introduceMe(name, age) {
+  return `Hello, my name is ${name} and I am ${age} years old`;
+}
 function getTotal(a, b) {
-  total = a ++ b;
+  total = a + b;
```
berkeli, line 11: This is missing a declaration, could you explain why it worked without a `const` or `let` declaration and why it would be a bad practice, even though it works?

mandatory/3-function-output.js
```diff
@@ -2,15 +2,18 @@
...
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
+/* This function takes two strings, word1 and word2, as arguments and concatenates them together using the concat() method. The concat() method returns a new string that is the result of joining two or more strings. */
 
 function concatenate(firstWord, secondWord, thirdWord) {
   // Write the body of this function to concatenate three words together.
   // Look at the test case below to understand what this function is expected to return.
+  return firstWord.concat(" ", secondWord.concat(" ", thirdWord));
```
berkeli, line 16: this works, but could you think of a solution using Template Literals (I think it will be a cleaner solution)

mandatory/4-tax.js
```diff
@@ -5,7 +5,10 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  price = price + price / 5;
+  return price;
+}
```
berkeli, line 11: See previous comment about pure functions

mandatory/4-tax.js
```diff
@@ -17,7 +20,10 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price) {
+  price = price + price / 5;
+  return `£${price.toFixed(2)}`;
+}
```
berkeli, line 26: Another one on pure functions

berkeli (changes requested):
Good job on the homework.

I have left a few comments to explore and think about, feel free to reach out on Slack or reply to comments if you would like to discuss them.

Oh and please try to get in the habit of filling out the pull request template, as this is very common in workplace.

extra/1-currency-conversion.js
nsaimk: only thing came my mind is "5.7" and "0.99" must have been const

mandatory/1-syntax-errors.js
nsaimk: i made it global variable by mistake, but has a side effect it is it may cause unexpected behaviour

mandatory/4-tax.js
nsaimk: Just found out what it is
   const newPrice = price + price/5

`...a pure function should not modify its input parameters or any other external variables or state`

berkeli: yes! that would be a better approach. We should try avoiding modification of parameters because if you call the same function multiple times you will keep getting different answers.

PR #515, opened 2023-02-21
london class 10-mahendra balal-javascript-week-1 

mabalal:

# Your Details

- Your Name:
- Your City:
- Your Slack Name:

# Homework Details

- Module:
- Week:

# Notes

- What did you find easy?

- What did you find hard?

- Any other notes?

### DIFF
--- a/mandatory/1-syntax-errors.js
+++ b/mandatory/1-syntax-errors.js
@@ -1,16 +1,17 @@
 // There are syntax errors in this code - can you fix it to pass the tests?
 
-function addNumbers(a b c) {
+function addNumbers(a, b, c) {
   return a + b + c;
 }
 
-function introduceMe(name, age)
-  return `Hello, my {name}` is "and I am $age years old`;
+function introduceMe(name, age) {
+  return `Hello, my name is ${name} and I am ${age} years old`;
+}
 
 function getTotal(a, b) {
-  total = a ++ b;
+  total = a + b;
 
-  return "The total is total";
+  return "The total is " + total;
 }
 
 /* 
--- a/mandatory/3-function-output.js
+++ b/mandatory/3-function-output.js
@@ -2,15 +2,17 @@
 function getRandomNumber() {
   return Math.random() * 10;
 }
-
+/*This function generates a random number and it will return with multiply by 10. */
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
+/* This function takes two strings, word1 and word2, as arguments and concatenates them together using the concat() method. The concat() method returns a new string that is the result of joining two or more strings. */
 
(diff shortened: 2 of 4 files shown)


mandatory/1-syntax-errors.js
```diff
@@ -1,16 +1,17 @@
...
 
-function introduceMe(name, age)
-  return `Hello, my {name}` is "and I am $age years old`;
+function introduceMe(name, age) {
+  return `Hello, my name is ${name} and I am ${age} years old`;
+}
 
 function getTotal(a, b) {
-  total = a ++ b;
+  total = a + b;
```
berkeli, line 12: This is missing a declaration here (`const` or `let`), can you think of why it still works? and why it would be a bad practice even though it works?

mandatory/3-function-output.js
```diff
@@ -2,15 +2,17 @@
...
 // Add comments to explain what this function does. You're meant to use Google!
 function combine2Words(word1, word2) {
   return word1.concat(word2);
 }
+/* This function takes two strings, word1 and word2, as arguments and concatenates them together using the concat() method. The concat() method returns a new string that is the result of joining two or more strings. */
 
 function concatenate(firstWord, secondWord, thirdWord) {
   // Write the body of this function to concatenate three words together.
   // Look at the test case below to understand what this function is expected to return.
+  return firstWord.concat(" ", secondWord.concat(" ", thirdWord));
```
berkeli, line 15: Can you think of a way doing this with template literals? I think it will be a cleaner solution

mandatory/4-tax.js
```diff
@@ -5,7 +5,10 @@
   Sales tax is 20% of the price of the product.
 */
 
-function calculateSalesTax() {}
+function calculateSalesTax(price) {
+  price = price + price / 5;
+  return price;
+}
```
berkeli, line 11: Here, you are modifying the argument that's passed to the function. This is usually a bad practice and should be avoided. 

If you would like to learn more, you can look into pure functions.

mandatory/4-tax.js
```diff
@@ -17,7 +20,10 @@ function calculateSalesTax() {}
   Remember that the prices must include the sales tax (hint: you already wrote a function for this!)
 */
 
-function addTaxAndFormatCurrency() {}
+function addTaxAndFormatCurrency(price) {
+  price = price + price / 5;
+  return `£${price.toFixed(2)}`;
+}
```
berkeli, line 26: See previous comment on pure functions

berkeli (approved):
Good work!

I left a few comments if you would like to improve the code and explore a few subjects that will come in handy.

mandatory/3-function-output.js
mabalal: okay thanks berkeli i will do that

